PR #1, opened 2026-03-11
add dockerfile and docker-compose

KhoaNguyen706:
We can run the backend on docker right now


KhoaNguyen706:
I should make a new branch

PR #4, opened 2026-03-13
Add Event structs

steeevin88:
pulled these from when I made an MVP, we can adjust these later, but it'll probably be easier to build our endpoints off of these


KhoaNguyen706 (approved):
I think it will help a lot.

PR #8, opened 2026-03-14
Add GET /events endpoint

miiiyyi:
Adds a GET /events endpoint
Changes:
- added GetEventsHandler in pkg/handlers/events.go
- wired GET /events route in cmd/server/main.go


steeevin88 (changes requested):
awesome stuff thanks for working on this 😃 had a few nits

PR #10, opened 2026-03-15
Add GET /events/:id endpoint

trista-chen-29:
Adds endpoint to retrieve a single event by ID.

Returns:
- 200 with event if found
- 404 if event does not exist

Tested locally using Postman.


cmd/server/main.go
```diff
@@ -32,6 +32,7 @@ func main() {
 
 	events := r.Group("/events")
 	{
+		events.GET("/:id", handlers.GetEventByIDHandler)
```
steeevin88: ah could we just call this `GetEventById` and omit the `Handler`? and could you change Mita's as well lol

just to be consistent with the POST endpoint (and the handler is a bit redundant anyway)

trista-chen-29: done!

PR #11, opened 2026-03-15
update CreateEvents function to write to mongo

codebyemily:


pkg/handlers/event.go
```diff
@@ -24,7 +27,7 @@ func GetEvents(c *gin.Context) {
 func GetEventByID(c *gin.Context) {
```
steeevin88: can we follow the pattern of the other handlers? where we call something in `/db`, ex:

```
if err := db.CreateEvent(&event); err != nil {
		c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to create event"})
		return
	}
```

this means the handler will be a lot simpler + the logic will be extracted to `/db/redis`

PR #14, opened 2026-03-24
add CORS Middleware

KhoaNguyen706:
Add CORS 
Feel free to lemme know which should i change.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -5,11 +5,14 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
 
 	"github.com/SCE-Development/SCEvents/pkg/db"
 	"github.com/SCE-Development/SCEvents/pkg/handlers"
 )
+
+const clientURL = "http://localhost:3000"
 
 func main() {
 	mongoURI := os.Getenv("MONGO_URI")
@@ -34,6 +37,12 @@ func main() {
 	}()
 
 	r := gin.Default()
+
+	config := cors.DefaultConfig()
+	config.AllowOrigins = []string{clientURL}
+	config.AllowCredentials = true
+	config.AddAllowHeaders("Authorization")
+	r.Use(cors.New(config))
 
 	r.GET("/ping", func(c *gin.Context) {
 		c.JSON(http.StatusOK, gin.H{
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,8 @@ go 1.25.5
 
 require (
 	github.com/gin-gonic/gin v1.12.0
+	github.com/redis/go-redis/v9 v9.7.0
+	github.com/gin-contrib/cors v1.4.0
 	go.mongodb.org/mongo-driver v1.17.9
 )
 


pkg/handlers/event.go
```diff
@@ -17,6 +18,9 @@ func GetEvents(c *gin.Context) {
 		})
 		return
 	}
+	if events == nil {
```
steeevin88: can we remove this? this is a good check but we should probably:
1. make it so that `db.GetEvents()` always returns valid JSON... so if there are no events, `GetEvents()` should return empty slices, instead of nil
2. we should move this to another PR... this PR is solely for CORS stuff

steeevin88: opened #15 for this, feel free to add it (should be a one-line change)

pkg/middleware/cors.go
```diff
@@ -0,0 +1,72 @@
+package middleware
```
steeevin88: i didn't do much research but i don't think we need to be our own CORS handler 😅 
packages like FastAPI and Gin should have their own built-in CORS handlers that we can just use, we should probably just import the necessarily functions and use them (as opposed to this custom handler)

i didn't do much reserach but Gemini says something like this could work, can we do this instead? and please test to make sure this works:

```
package main

import (
	"time"

	"github.com/gin-contrib/cors"
	"github.com/gin-gonic/gin"
	// ... other imports
)

func main() {
	r := gin.New()

	// The clean, 10-line equivalent of the custom middleware
	r.Use(cors.New(cors.Config{
		AllowOrigins:     []string{"http://localhost:3000", "http://127.0.0.1:3000"},
		AllowMethods:     []string{"GET", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"},
		AllowHeaders:     []string{"Origin", "Content-Type", "Authorization", "Accept"},
		ExposeHeaders:    []string{"Content-Length"},
		AllowCredentials: true,
		MaxAge:           12 * time.Hour,
	}))

	r.Use(gin.Logger())
	r.Use(gin.Recovery())

    // ... rest of your routes
}
```

Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN go mod download
 
 COPY cmd/ cmd/
 COPY pkg/ pkg/
-RUN go build -o ./scevents ./cmd/server
+RUN go mod tidy && go build -o ./scevents ./cmd/server
```
steeevin88: remove this; we should be running go mod tidy locally and not in the Docker container

i will let everyone else know this but thanks for noticing this 😅

steeevin88 (commented):
thanks for the changes, let me know if anything is confusing 👍 

please update the PR title + description to reflect all the changes after, thanks!

Dockerfile
steeevin88: wait again can we remove go mod tidy from Dockerfile? we should just be running `go mod tidy` locally, the Docker container shouldn't have this responsibility

cmd/server/main.go
```diff
@@ -22,32 +24,41 @@ func main() {
...
-		events.POST("/", handlers.CreateEvent)
-		events.DELETE("/:id", handlers.DeleteEventByID)
-	}
+	r.GET("/events", handlers.GetEvents)
+	r.POST("/events", handlers.CreateEvent)
+	r.GET("/events/:id", handlers.GetEventByID)
+	r.DELETE("/events/:id", handlers.DeleteEventByID)
 
-	r.Run()
+	// Default :8080 — must match docker-compose `ports: "8002:8080"` (host:container).
```
steeevin88: we don't need this comment

let's also open another PR after this PR is merged that changes the default port of the server to 8082 instead of 8080, such that if we just ran the server alone it would run on port 8080. right now, the server default to port 8080, and with your change we expose the server (through the Docker container) to our local port 8082.

and let's move that change into a separate PR so this current PR focuses solely on the CORS stuff, thanks!

PR #21, opened 2026-04-01
Write event headcount to Redis on creation 

codebyemily:
merged with #19 
issue #17 

validate that the keys in Redis are being populated 
image

persist if the containers are shut down
image


pkg/db/redis.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/go-redis/redis/v9"
+	"github.com/redis/go-redis/v9"
```
steeevin88: not sure what changed here 😅  do you have an auto-formatter?

could you run `go mod tidy`? i think we'll need to run this if we ever make changes to any `.go` files

or if these changes are caused by running `go mod tidy` then please ignore oops

pkg/handlers/event.go
```diff
@@ -58,4 +58,21 @@ func CreateEvent(c *gin.Context) {
 	c.JSON(http.StatusCreated, createdEvent)
 }
```
steeevin88: almost done; you'll need to call the `SetEventHeadcount` in the CreateEvent handler, i think after creating the event in MongoDB:

https://github.com/SCE-Development/SCEvents/blob/7c75042dc7165e60b39edf94319e6a309aec1e56/pkg/handlers/event.go#L56

steeevin88 (changes requested):
thanks for the quick PR!

steeevin88 (approved):
awesome, thanks! let's wait til cindy merges her PR, then let's rebase this PR before merging

Issue #27, opened 2026-04-08
Implement event registration endpoint in SCEvents


maernest04:
Add server-side support in SCEvents for event registration via POST /events/:id/register, including payload validation and duplicate registration protection. Kafka producer has not been created so this issue will be solved with a shell template while we wait for the producer to be made

- define RegistrationPayload → what do we expect when someone registers?
- define RegisterForEvent function
- ensure payload is valid JSON
- Check for duplicate registration

PR #28, opened 2026-04-08
Added event registration endpoint

maernest04:
For [issue](https://github.com/SCE-Development/SCEvents/issues/27)

This pull adds SCEvents registration payload validation and duplicate checks and wires Clark event registration form to submit to the backend.

### DIFF
--- a/pkg/db/redis.go
+++ b/pkg/db/redis.go
@@ -56,6 +56,10 @@ func EventHeadcountKey(eventID string) string {
 	return fmt.Sprintf("event:%s:headcount", eventID)
 }
 
+func EventRegistrantsKey(eventID string) string {
+	return fmt.Sprintf("event:%s:registrants", eventID)
+}
+
 func SetEventHeadcount(eventID string, capacity int) error {
 	if redisClient == nil {
 		return fmt.Errorf("redis not connected")
@@ -131,4 +135,17 @@ func ReleaseEventSeat(eventID string) error {
 	key := EventHeadcountKey(eventID)
 
 	return redisClient.Incr(ctx, key).Err()
+}
+
+func IsUserRegisteredForEvent(eventID string, userID string) (bool, error) {
+	if redisClient == nil {
+		return false, fmt.Errorf("redis not connected")
+	}
+	ctx := context.Background()
+	key := EventRegistrantsKey(eventID)
+	isMember, err := redisClient.SIsMember(ctx, key, userID).Result()
+	if err != nil {
+		return false, err
+	}
+	return isMember, nil
 }
--- a/pkg/event/registration.go
+++ b/pkg/event/registration.go
@@ -1,0 +1,61 @@
+package event
+
+import (
+	"strings"
+	"time"
+)
+
+type RegistrationFormValidationError struct {
+	Field   string
+	Message string
+}
+
+func (e *RegistrationFormValidationError) Error() string {
+	return e.Message
+}
+
+type Registrant struct {
+	Name   string `json:"name" binding:"required"`
+	Email  string `json:"email" binding:"required,email"`
+	UserID string `json:"user_id,omitempty"`
+}
+
+type RegistrationPayload struct {
+	Registrant              Registrant      `json:"registrant" binding:"required"`
+	RegistrationFormAnswers map[string]any  `json:"registration_form_answers" binding:"required"`
+	Timestamp               time.Time       `json:"timestamp"`
+}
+
+func (e *Event) ValidateRegistration(answers map[string]any) error {
+	for _, question := range e.RegistrationForm {
+		if !question.Required {
+			continue
+		}
+
+		answer, exists := answers[question.ID]
+		if !exists {
+			return &RegistrationFormValidationError{
+				Field:   question.ID,
+				Message: "missing required answer",
+			}
+		}
+
+		switch v := answer.(type) {
+		case string:
+			if strings.TrimSpace(v) == "" {
+				return &RegistrationFormValidationError{
+					Field:   question.ID,
+					Message: "required answer cannot be empty",
+				}
+			}
+		case []any:
+			if len(v) == 0 {
+				return &RegistrationFormValidationError{
+					Field:   question.ID,
+					Message: "required answer cannot be empty",
+				}
+			}
+		}
+	}
+	return nil
+}
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -1,8 +1,9 @@
 package handlers
 
 import (
+	"errors"
 	"net/http"
-
+	"strings"
 	"github.com/SCE-Development/SCEvents/pkg/db"
 	event "github.com/SCE-Development/SCEvents/pkg/event"
 	"github.com/gin-gonic/gin"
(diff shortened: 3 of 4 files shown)


pkg/db/redis.go
```diff
@@ -64,3 +68,25 @@ func SetEventHeadcount(eventID string, capacity int) error {
...
+	ctx := context.Background()
+	key := EventRegistrantsKey(eventID)
+	isMember, err := redisClient.SIsMember(ctx, key, userID).Result()
+	if err != nil {
+		return false, err
+	}
+	return isMember, nil
+}
+
+func AddUserRegistration(eventID string, userID string) error {
```
steeevin88: let's remove this; good proactivity but for now let's assume that our endpoint doesn't actually register anyone. i think the "isUserRegistered" check is really good so we can keep that, but in our eventual workflow once all our checks pass, we will pass the registration request to our producer routine to "handle registration", so let's not handle registration here

pkg/event/event.go
```diff
@@ -30,3 +30,15 @@ type Event struct {
...
+type Registrant struct {
+	Name   string `json:"name" binding:"required"`
+	Email  string `json:"email" binding:"required,email"`
+	UserID string `json:"user_id,omitempty"`
+}
+
+type RegistrationPayload struct {
+	Registrant  Registrant      `json:"registrant" binding:"required"`
+	Answers     map[string]any  `json:"answers" binding:"required"`
+	SubmittedAt string          `json:"submitted_at,omitempty"`
```
steeevin88: can we do this instead:
```go
Timestamp time.Time      `json:"timestamp"`
```

pkg/event/event.go
```diff
@@ -30,3 +30,15 @@ type Event struct {
...
+
+type Registrant struct {
+	Name   string `json:"name" binding:"required"`
+	Email  string `json:"email" binding:"required,email"`
+	UserID string `json:"user_id,omitempty"`
+}
+
+type RegistrationPayload struct {
+	Registrant  Registrant      `json:"registrant" binding:"required"`
+	Answers     map[string]any  `json:"answers" binding:"required"`
```
steeevin88: can we call this `FormAnswers` instead? or maybe `RegistrationFormAnswers`? just to be more specific that the answers are tied to the existing types, maybe we can connect them better somehow

pkg/event/event.go
```diff
@@ -30,3 +30,15 @@ type Event struct {
...
+	SubmittedAt string          `json:"submitted_at,omitempty"`
+}
```
steeevin88: can we move these types to a new `registration.go`?

pkg/handlers/event.go
```diff
@@ -82,4 +82,104 @@ func DeleteEventByID(c *gin.Context) {
...
+		return
+	}
+
+	contentType := c.GetHeader("Content-Type")
+	if !strings.Contains(strings.ToLower(contentType), "application/json") {
+		c.JSON(http.StatusBadRequest, gin.H{
+			"error": "content type must be application/json",
+		})
+		return
+	}
```
steeevin88: i think these lines aren't necessary, `c.ShouldBindJSON` would catch this

pkg/handlers/event.go
```diff
@@ -82,4 +82,104 @@ func DeleteEventByID(c *gin.Context) {
...
+				c.JSON(http.StatusBadRequest, gin.H{
+					"error": "required answer cannot be empty",
+					"field": question.ID,
+				})
+				return
+			}
+		}
+	}
+
+	alreadyRegistered, err := db.IsUserRegisteredForEvent(eventID, payload.Registrant.UserID)
```
steeevin88, line 210: can we move this check BEFORE the form validation? it makes more sense to do this super quick check first before doing the potentially long form validation (where we have to grab the event, then extract it's form questions, and check for each question, does it match the registration request's form answers)

pkg/handlers/event.go
```diff
@@ -82,4 +82,104 @@ func DeleteEventByID(c *gin.Context) {
...
+		return
+	}
+	if alreadyRegistered {
+		c.JSON(http.StatusConflict, gin.H{
+			"error": "user is already registered for this event",
+		})
+		return
+	}
+
+	if err := db.AddUserRegistration(eventID, payload.Registrant.UserID); err != nil {
```
steeevin88: similar to earlier comment, let's remove this (since we don't want to actually handle registration here)

pkg/handlers/event.go
```diff
@@ -82,4 +82,104 @@ func DeleteEventByID(c *gin.Context) {
...
+		c.JSON(http.StatusInternalServerError, gin.H{
+			"error": "failed to store registration",
+		})
+		return
+	}
+
+	c.JSON(http.StatusAccepted, gin.H{
```
steeevin88, line 247: let's change this message to "registration request sent" instead... since even if a user sends the registration request, they aren't necessarily guaranteed to be granted registration to the event

ex. if millions of people registered for one last spot, basically all of them may be able to send a registration request, but not all of them will be accepted

pkg/handlers/event.go
```diff
@@ -82,4 +82,104 @@ func DeleteEventByID(c *gin.Context) {
...
+
+	event, err := db.GetEventByID(eventID)
+	if err != nil {
+		c.JSON(http.StatusNotFound, gin.H{
+			"error": "event not found",
+		})
+		return
+	}
+
+	for _, question := range event.RegistrationForm {
```
steeevin88: can we extract this into a helper? ex. `validateFormAnswers()`

it should probably live elsewhere; one approach is to attach the method to `pkg/event/registration.go` once you've created it

i asked gemini 😅 but ex.
```
// pkg/event/registration.go
package event // (or whatever the package name is declared as in this folder)

import (
	"fmt"
	"strings"
)

// ... (Your Registrant and RegistrationPayload structs go here) ...

// ValidateRegistration checks if the provided answers satisfy the event's required form questions.
func (e *Event) ValidateRegistration(answers map[string]any) error {
	for _, question := range e.RegistrationForm {
		if !question.Required {
			continue
		}

		answer, exists := answers[question.ID]
		if !exists {
			return fmt.Errorf("missing required answer for field: %s", question.ID)
		}

		switch v := answer.(type) {
		case string:
			if strings.TrimSpace(v) == "" {
				return fmt.Errorf("required answer cannot be empty for field: %s", question.ID)
			}
		case []any:
			if len(v) == 0 {
				return fmt.Errorf("required answer cannot be empty for field: %s", question.ID)
			}
		}
	}
	return nil
}
```


and then our final method would look a lot cleaner

steeevin88 (changes requested):
awesome stuff overall, left some comments to ideally re-order logic + remove actual registration handling since we'll leave that to the producer

had gemini generate this but ideally the method will look something like this (maybe without the AI comments lol):

```
func RegisterForEvent(c *gin.Context) {
	eventID := c.Param("id")
	if eventID == "" {
		c.JSON(http.StatusBadRequest, gin.H{
			"error": "event id is required",
		})
		return
	}

	var payload types.RegistrationPayload
	if err := c.ShouldBindJSON(&payload); err != nil {
		c.JSON(http.StatusBadRequest, gin.H{
			"error": "invalid JSON payload",
		})
		return
	}

	if strings.TrimSpace(payload.Registrant.UserID) == "" {
		c.JSON(http.StatusUnauthorized, gin.H{
			"error": "login required to register for event",
		})
		return
	}

	// Fail Fast: Check if already registered
	alreadyRegistered, err := db.IsUserRegisteredForEvent(eventID, payload.Registrant.UserID)
	if err != nil {
		c.JSON(http.StatusInternalServerError, gin.H{
			"error": "failed to verify existing registration",
		})
		return
	}
	if alreadyRegistered {
		c.JSON(http.StatusConflict, gin.H{
			"error": "user is already registered for this event",
		})
		return
	}

	// Fetch Event and Validate Answers
	event, err := db.GetEventByID(eventID)
	if err != nil {
		c.JSON(http.StatusNotFound, gin.H{
			"error": "event not found",
		})
		return
	}

	if err := event.ValidateRegistration(payload.Answers); err != nil {
		c.JSON(http.StatusBadRequest, gin.H{
			"error": err.Error(),
		})
		return
	}

	// once here, we would run some producer routine function to begin processing the registration request

	c.JSON(http.StatusAccepted, gin.H{
		"message":  "registration request sent",
		"event_id": eventID,
	})
}
```

pkg/event/registration.go
```diff
@@ -0,0 +1,61 @@
...
+	Name   string `json:"name" binding:"required"`
+	Email  string `json:"email" binding:"required,email"`
+	UserID string `json:"user_id,omitempty"`
+}
+
+type RegistrationPayload struct {
+	Registrant              Registrant      `json:"registrant" binding:"required"`
+	RegistrationFormAnswers map[string]any  `json:"registration_form_answers" binding:"required"`
+	Timestamp               time.Time       `json:"timestamp"`
+}
```
steeevin88, line 27: @KhoaNguyen706 i think these are technically our DTOs lol, these are types that validate that the data transferred from the frontend is actually what we want

steeevin88 (approved):
lgtm but let's rebase, it says there are merge conflicts but not sure what the changes will be

let's also:
- close #27 
- work on #34 after

PR #29, opened 2026-04-09
Add registration persistence models and Mongo lifecycle helpers

trista-chen-29:
## Summary
Adds the registration persistence layer for async event registration processing.

## Changes
- add `pkg/registration/registration.go`
  - registration status and decision reason types
  - submit request model
  - persisted registration request model
  - Kafka registration message model
- add `pkg/db/registration.go`
  - create pending registration
  - get registration by ID
  - check for existing accepted registration
  - mark registration accepted
  - mark registration rejected

## Validation
- ran `gofmt`
- ran `go build ./...`
- ran `go test ./...`

## Notes
- This PR only covers the data model + persistence side.
- Consumer (Kafka processing) logic will be added in a follow-up PR

### DIFF
--- a/pkg/db/registration.go
+++ b/pkg/db/registration.go
@@ -1,0 +1,148 @@
+package db
+
+import (
+	"context"
+	"time"
+
+	"github.com/SCE-Development/SCEvents/pkg/registration"
+	"go.mongodb.org/mongo-driver/bson"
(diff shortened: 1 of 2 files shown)


steeevin88 (approved):
lgtm + thanks for adding this, did not think of this while planning but we totally should persist registrations

PR #30, opened 2026-04-09
Fix event package name from types to event

trista-chen-29:
## Summary
Updates the `pkg/event` package name from `types` to `event` for consistency with imports and usage across the codebase.

## Changes
- rename package declaration in `pkg/event/event.go`
- update handler usage to reference `event.Event`

## Validation
- ran `gofmt`
- ran `go build ./...`
- ran `go test ./...`

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -1,4 +1,4 @@
-package types
+package event
 
 type AnswerDetails struct {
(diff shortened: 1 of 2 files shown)


steeevin88 (approved):
awesome, thanks for finding this 👍

Issue #35, opened 2026-04-11
Allow updating events


steeevin88:
totally missed this in our initial endpoints (but to be fair this is more complicated + required our types to be a bit more flushed out)

when an event is created, it's reasonable to assume that an admin will want to be able to edit the event details (ex. change the event location, update the date, remove attendees, increase headcount)

create the logic for this... there's a few points to consider:
1. we'll need new endpoints for this
2. we need to ensure that only the APPROPRIATE people are allowed to edit the event --> admin or global user??
3. we'll need to update the event details in MongoDB
4. obviously, event headcounts are stored in Redis... we'll need to add logic to support that as well (updating the headcount, especially if it's already been modified)
5. we'll need to create the frontend experience to support updating the event (again, only allow this if the user is an admin of the event

please split this up into multiple PRs... ex. one creating the endpoint, one that updates headcount in redis appropriately, etc...

steeevin88:
let's close this once we've verified that it works properly through the frontend

steeevin88:
merged https://github.com/SCE-Development/Clark/pull/2080

PR #40, opened 2026-04-11
add Update Event endpoint

KhoaNguyen706:
Issue: [35](https://github.com/SCE-Development/SCEvents/issues/35)

- Add the endpoints

For my previous projects, I usually JWT to protect and authorize the endpoints. 
- If we dont need to use JWT, Clark will have us authorization and authentication ?

### DIFF
--- a/pkg/db/event.go
+++ b/pkg/db/event.go
@@ -87,26 +87,16 @@ func DeleteEventByID(id string) error {
 	return nil
 }
 
-func UpdateEventByID(id string, e event.Event) error {
+// UpdateEventByID performs a partial update on an event document.
+// Only fields present in the provided map are updated via $set.
+func UpdateEventByID(id string, fields map[string]interface{}) error {
 	coll := GetEventsCollection()
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	// Convert the event struct to a map for $set update
-	// We want to avoid overwriting _id and created_at if they are not intended to be changed
 	update := bson.M{
-		"$set": bson.M{
-			"name":              e.Name,
-			"date":              e.Date,
-			"time":              e.Time,
-			"location":          e.Location,
-			"description":       e.Description,
-			"admins":            e.Admins,
-			"registration_form": e.RegistrationForm,
-			"max_attendees":     e.MaxAttendees,
-			"status":            e.Status,
-		},
+		"$set": fields,
 	}
 
 	result, err := coll.UpdateOne(ctx, bson.M{"_id": id}, update)
--- a/pkg/db/event.go
+++ b/pkg/db/event.go
@@ -86,3 +86,37 @@ func DeleteEventByID(id string) error {
 
 	return nil
 }
+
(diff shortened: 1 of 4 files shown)


pkg/db/event.go
```diff
@@ -86,3 +86,37 @@ func DeleteEventByID(id string) error {
...
+			"date":              e.Date,
+			"time":              e.Time,
+			"location":          e.Location,
+			"description":       e.Description,
+			"admins":            e.Admins,
+			"registration_form": e.RegistrationForm,
+			"max_attendees":     e.MaxAttendees,
+			"status":            e.Status,
+		},
+	}
```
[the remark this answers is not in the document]
steeevin88: this is a good shout maybe let's do this?

steeevin88 (changes requested):
hmm... i think for auth we should just do a basic check: "if user is an admin of the event, they can edit it"

can you add this for now?

eventually we will need to modify this though... in terms of Clark membership levels, we should allow all admins to edit any event, and officers can edit IF they are an admin of the events...

https://github.com/SCE-Development/Clark/blob/ea70088c8baee4cd2caa8c5d6d0ddfa3f1cd481b/src/Enums.js#L23-L30

this might be a little too complicated but we can worry about this later, opened #41 for this. let's work on it after you finish #35

steeevin88:
also please link the issue in this pr description next time thanks 😄

pkg/handlers/event.go
```diff
@@ -81,3 +82,82 @@ func DeleteEventByID(c *gin.Context) {
...
+	}
+
+	// Check if user is an admin of this event
+	isAdmin := false
+	for _, admin := range existingEvent.Admins {
+		if admin == userID {
+			isAdmin = true
+			break
+		}
+	}
```
steeevin88: let's make this a helper in `pkg/event/event.go`

pkg/handlers/event.go
```diff
@@ -81,3 +82,82 @@ func DeleteEventByID(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, gin.H{
+			"error": "invalid JSON payload",
+		})
+		return
+	}
+
+	// Strip immutable fields that should not be overwritten
+	delete(fields, "id")
+	delete(fields, "_id")
+	delete(fields, "created_at")
```
steeevin88: let's also make this a helper in `pkg/event/event.go`, ex. `SanitizeUpdateFields`

pkg/handlers/event.go
steeevin88: then, this part could look simpler:

```
if !existingEvent.IsAdmin(userID) {
	c.JSON(http.StatusForbidden, gin.H{"error": "you are not an admin of this event"})
	return
}
```

steeevin88 (changes requested):
yes you'll be adding a frontend to support this

I'm imagining an edit button that renders if you're an admin of an event, and clicking on the edit button will move you to an edit page or something

you can decide what the UI will be

steeevin88 (approved):
lgtm, once the frontend is created we can better debug this + if fixes are needed we can make a new PR

PR #42, opened 2026-04-12
Update Headcount logic for the Update endpoint

KhoaNguyen706:
Issue [35](https://github.com/SCE-Development/SCEvents/pull/40l)
1. First commit:
- add Headcount redis logic
- We can edit Headcount logic now!!

2. Second commit:
- change the math logic 
- the formula: The math: new_redis = new_max - (original_max - current_redis)
- How it works ? I think we get current Redis headcount if Admin edit the seat, and then I will do math and set HeadCount again. The example, Admin of event edit max seat from 100 to 110, and  current redis headcount is 70 .  new_max = 110, original max = 100,  current_reddis = 70.. New redis headcount that we will set  is 40 due to formula. It also works if admin of events decrease max_seat. new max = 90 , new redis = 20.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -56,6 +56,7 @@ func main() {
 		events.GET("/:id", handlers.GetEventByID)
 		events.POST("/", handlers.CreateEvent)
 		events.DELETE("/:id", handlers.DeleteEventByID)
+		events.PUT("/:id", handlers.UpdateEventByID)
 	}
 
 	r.Run(":8002")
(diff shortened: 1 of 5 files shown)


steeevin88 (changes requested):
a few things:

1. if the event headcount was decremented already, we need to account for that... not sure what the best way to do this would be. i think you'd need to grab the ORIGINAL headcount (which should be in event entry in Mongo?) and subtract the current headcount in Redis from that number to get the number of seats that we've already given out. then, when you set the new value, decrement the new value by the seats that we've already given
2. this PR is branched atop your other PR (#40). let's merge #40 into main first and then merge #42 into main (as opposed to merging this PR into your branch)

steeevin88 (approved):
lgtm, merge this into main after #40 please

PR #44, opened 2026-04-13
Fix import cycle and centralize models into pkg/models

trista-chen-29:
## Summary
This PR restructures the project by centralizing shared event and registration models into a single `pkg/models` package. It also resolves the previous import cycle in the registration flow.

## Changes
- moved event models (`Event`, `FormQuestion`, etc.) into `pkg/models`
- moved registration models (`RegistrationRequest`, `RegistrationPayload`, etc.) into `pkg/models`
- moved event-related methods (`IsAdmin`, `ValidateRegistration`) alongside the `Event` model
- removed `pkg/event` package
- updated `pkg/db`, `pkg/handlers`, and `pkg/registration` to use `pkg/models`
- updated `go.mod` to reflect direct Kafka dependency

## Result
- resolves the import cycle between `db` and `registration`
- provides a cleaner and more scalable structure for shared domain types
- keeps business logic (handlers, DB, Kafka consumer) separate from shared models

## Notes / Future Improvements
- no functional changes; this is a structural refactor only
- Kafka consumer and DB logic remain unchanged


steeevin88 (changes requested):
really awesome catch, thanks for being proactive (again) and making a PR for this

i dislike the folder strucutre that this would introduce though, we should definitely follow what you mentioned in your PR description where we unify everything into a single `pkg/models` directory or something similar; could you do that with this PR instead?

you'll probably need to update a few things but that's fine, thanks

steeevin88 (approved):
am trusting you that nothing broke 😄

Issue #32, opened 2026-04-11
Create Config struct


steeevin88:
`main.go` is getting a bit messy with all the environment variables needed (ex. `os.Getenv("MONGO_URI")`)... however, they're completely necessary. additionally, currently we hardcode the server's port to 8002; this should also be an environment variable as well

https://github.com/SCE-Development/SCEvents/blob/279d801bce467c72547041a15fcc44a7723aa552/cmd/server/main.go#L61

to make this cleaner, let's create a configuration loader

- [ ] Create a new `internal/config/config.go` package (or place it in `cmd/server/config.go`) --> anything in `/internal` shouldn't be accessible to other applications... which is what we want here
- [ ] Define an `AppConfig` struct that holds all typed environment variables (Mongo URI, Redis Addr, Kafka config, Port etc.); ensure there are defaults...
- [ ] Create a `Load()` function that reads from the environment, handles defaults, and returns the struct.
- [ ] Refactor `main.go` to call `config.Load()` once at startup and pass the specific fields down to our handlers and consumers.
- [ ] Update the `docker-compose.yml` to pass in the new port environment variable 

be sure to update #31 depending on if this issue is merged before/after

PR #47, opened 2026-04-13
Add centralized config loader for server settings

miiiyyi:
#32 
- added centralized config loader in internal/config/config.go
- moved environment variable loading out of main.go
- made server port configurable through SERVER_PORT
- updated docker-compose.yml to pass the new port environment variable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN go mod download
 
 COPY cmd/ cmd/
 COPY pkg/ pkg/
+COPY internal/ internal/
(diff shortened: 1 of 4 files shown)


steeevin88 (approved):
awesome thanks! make sure to close the issue after, not sure if it auto-closes

weslayer (commented):
i think these env vars should use the ${} format because what's the point of getting rid of the hardcoding in the real code if you hardcode it in the docker compose file right 🤔 this would require creating a .env if you'd like, but the ${} format allows defaults as well

Issue #34, opened 2026-04-11
Create + write registration requests to MongoDB


steeevin88:
#29 adds the logic to persist registrations; this is super important as when a user register's, we basically send their request off to Kafka, and if there is space then their request may be processed and they're given a spot (successful registration). however, this flow means that there isn't a request that is sent to the user that tells them if they got a spot in the event or not (if their registration was accepted)

to do this... we're going to have the frontend POLL of status of their registrations... which are stored in MongoDB (again, through the logic in #29)

#28 adds the registration endpoint, but we should probably create the registration request object and write it to MongoDB here (in addition to enqueue the request to the producer script)... so let's do that

- [ ] update the registration endpoint to create + write to MongoDB

most of the heavy lifting should already be done via #29

Issue #38, opened 2026-04-11
Create Producer Routine


steeevin88:
#31 adds the consumer routine that reads RegistrationRequests from Kafka... but we need the producer routine that actually writes the requests to Kafka (the `registrations` topic)

- [ ] Create pkg/registration/producer.go with the Producer struct and NewProducer constructor.
- [ ] Implement the PublishRegistration method. Ensure the payload structure matches what the Consumer expects (specifically including the Mongo RequestID, EventID, and UserID).
- [ ] Wire up the Producer in main.go (pass it the brokers/topic from our AppConfig) and pass the instance down to the HTTP handlers.
- [ ] Update the POST /events/:id/register endpoint to execute the "Outbox" flow:
1. Write Pending registration to MongoDB.
2. Call producer.PublishRegistration with the generated RequestID.
3. Return HTTP 202 Accepted to the client.
- [ ] Add producer.Close() to the graceful shutdown block in main.go.

let's also ensure we add the async option in the kafka writer config

please be sure to test this 😄

PR #52, opened 2026-04-14
Created mongo registration requests functions

maernest04:
This PR is for issue #34 which is creating write registration to MongoDB. The code will take registration payloads and write them to MongoDB.

## Changes made
- RegisterForEvent updated to create and save a RegistrationRequest that will be written to MongoDB
- GetRegistrationStatus was added to allow for polling of a user's registration, basically live status of whether they registered or not (see below image)
- Created the Kafka send off function for the producer (producer code doesn't currently exist)
- registration.go uses multiple resources like publish_state and publish_attempts for automatic retrying of registration
- HasPendingOrAcceptedRegistration was added to prevent users from sending many requests, users will instead see an unclickable button (see below image)

## Images

Screenshot 2026-04-14 at 12 16 17 AM
Screenshot 2026-04-14 at 12 16 43 AM

## Things to note
- For testing, I had to make some changes to the front end to call my new functions. These changes have not been committed or had a PR in Clark.
- For SCEvents, I had to create a dummy producer.go file inside the pkg/registration folder. This file was not included in this commit and PR.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -35,6 +35,20 @@ func main() {
 		}
 	}()
 
+	kafkaBroker := os.Getenv("KAFKA_BROKER")
+	kafkaTopic := os.Getenv("KAFKA_TOPIC")
+	kafkaGroupID := os.Getenv("KAFKA_GROUP_ID")
+
+<<<<<<< HEAD
+<<<<<<< HEAD
+=======
+	if kafkaBroker == "" || kafkaTopic == "" {
+		log.Println("warning: KAFKA_BROKER or KAFKA_TOPIC empty")
+	}
+
+>>>>>>> c493044 (Removed kafka implementation stuff)
+=======
+>>>>>>> 0e55751 (Removing code based off Steven's comments)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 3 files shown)


steeevin88 (changes requested):
hmm, looks like PR is doing a lot of what #38 is supposed to do; in particular, the producer goroutine should be writing to Kafka, not the registration endpoint

could you strip out this logic, i think there's only a few spots that are extra? might be helpful to keep it somewhere in case you need it again in #38 , but for the context of this PR we should just be focused on the DB persistence

steeevin88:
ex. 

https://github.com/SCE-Development/SCEvents/blob/b2b28d8296e16d5f0ce47bcaf8eecbadd0bf2639/cmd/server/main.go#L66

i could be mistaken but i don't think this exists in our codebase right? if we merge this PR we'll break the service

cmd/server/main.go
```diff
@@ -35,21 +38,29 @@ func main() {
 		}
 	}()
 
+	kafkaBroker := os.Getenv("KAFKA_BROKER")
+	kafkaTopic := os.Getenv("KAFKA_TOPIC")
+	kafkaGroupID := os.Getenv("KAFKA_GROUP_ID")
+
+	if kafkaBroker == "" || kafkaTopic == "" {
```
steeevin88: let's remove this, i'm pretty with our (new) config setup there will always be values?

pkg/handlers/event.go
```diff
@@ -314,8 +246,84 @@ func RegisterForEvent(c *gin.Context) {
...
+
+	_, err = db.CreatePendingRegistration(mongoRequest)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{
+			"error": "failed to persist registration request",
+		})
+		return
+	}
+
+	// Notice: Kafka Publisher hook will later go here in Issue #38.
```
steeevin88: let's remove this comment

steeevin88 (approved):
looks fine to me, if the most recent changes didn't break logic then no need to test + feel free to merge

obviously rebase + mark the below comments as resolved before merging, thanks

PR #53, opened 2026-04-14
Fix empty event _id generation in CreateEvent

### LINKED ISSUE #49: Disallow empty event ID
I was able to verify the visibility validation logic on the backend with the following observations:

* When creating an event without explicitly setting these fields, the backend correctly defaults to `status = draft` and `visibility = public`.
* When attempting to create a private event (`visibility = private`) without providing `minimum_visible_role`, the request is rejected as expected by the validation logic.

However, I was not able to fully test the PATCH/update flows end to end due to a separate issue I encountered in `CreateEvent()`:

* Newly created events are being inserted into MongoDB with an empty string as the `_id` (i.e., `_id: ""`).
* This causes subsequent operations (such as creating additional events or updating existing ones via PATCH) to fail, since the ID is not valid or unique.
* I confirmed this directly in MongoDB using:

  ```js
  db.events.find({ _id: "" }).pretty()
  ```

  which returns the inserted event with an empty `_id`.

Because of this, I couldn’t reliably test updating visibility (e.g., switching between public/private or validating `minimum_visible_role` on PATCH), since the event IDs are not usable.

This issue appears to be independent of the visibility validation changes in this PR and likely needs to be addressed separately (e.g., ensuring `_id` is properly generated on insert)?

_Originally posted by @trista-chen-29 in https://github.com/SCE-Development/SCEvents/issues/45#issuecomment-4238333247_

we should add a fix to this

trista-chen-29:
Closes #49

## Summary
Ensure that new events are created with a valid non-empty `_id`.

## Fix
Generate a string ID (`primitive.ObjectID().Hex()`) before calling `InsertOne()` when `e.ID` is empty.

## Verification
Example terminal result after fix:

```bash
curl -X POST http://localhost:8002/events/ \
  -H "Content-Type: application/json" \
  -d '{
    "name": "Test Event With Status",
    "date": "2026-04-20",
    "time": "18:00",
    "location": "Student Union",
    "description": "Manual status test",
    "admins": ["user123"],
    "registration_form": [],
    "max_attendees": 50,
    "status": "draft"
  }'
```

```json
{"id":"69ddefd15aeab4a1f0851d92","name":"Test Event With Status","date":"2026-04-20","time":"18:00","location":"Student Union","description":"Manual status test","admins":["user123"],"registration_form":[],"max_attendees":50,"created_at":"","status":"draft"}
```

Previously, events could be inserted with an empty string `_id`, which caused later create/update operations to fail.

### DIFF
--- a/pkg/db/event.go
+++ b/pkg/db/event.go
@@ -74,19 +74,13 @@ func CreateEvent(e models.Event) (*models.Event, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	res, err := coll.InsertOne(ctx, e)
-	if err != nil {
-		return nil, err
+	if e.ID == "" {
+		e.ID = primitive.NewObjectID().Hex()
 	}
 
-	// if Mongo generated an ID, reflect it back in the event
-	if e.ID == "" {
-		switch id := res.InsertedID.(type) {
-		case primitive.ObjectID:
-			e.ID = id.Hex()
-		case string:
-			e.ID = id
-		}
+	_, err := coll.InsertOne(ctx, e)
+	if err != nil {
+		return nil, err
 	}
 
 	return &e, nil


steeevin88 (approved):
lgtm, thanks!

PR #55, opened 2026-04-15
Use redis:alpine image

steeevin88:
**CHANGES**
- changed redis image to be smaller

**TESTING**
- i didn't test but surely it doesn't break anything


c1ndytat (approved):
molly tea

PR #58, opened 2026-04-15
Add simple Go CI workflow

### LINKED ISSUE #46: Create simple CI
#44 solves an issue where we had an import cycle

during development, we should have caught this locally by running something like `go build`... to improve developer experience let's just make a simple Github action to do this for us. i'm not sure what the exact `go` command is though

this task should be pretty copilot-able lol but what i'm imagining is we should check for two things:
- verify that dependencies are synced + properly updated --> `go mod tidy`, then `go mod vendor`, then `git diff --exit-code` to see if files changed
- can we still build --> `go build`

once we merge the CI in main, i'll update this github repo to make it merge-blocking

feel free to add/remove steps as you deem necessary

miiiyyi:
- added a simple GitHub Actions workflow for Go CI
- verifies go.mod and go.sum stay in sync after go mod tidy
- builds all packages with go build ./...
- ignores vendor/ since the repo is not tracking vendored dependencies
Closes #46

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -1,4 +1,4 @@
-name: Go CI
+name: Golang Validation
 
 on:
   pull_request:
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -1,0 +1,33 @@
+name: Go CI
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+jobs:
+  verify-and-build:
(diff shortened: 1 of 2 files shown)


.github/workflows/go-ci.yml
```diff
@@ -0,0 +1,33 @@
+name: Go CI
```
steeevin88: just being nitpicky, could we call it `Golang Validation`

steeevin88 (approved):
looks like someone didn't run `go mod tidy`  😭 we can do that in a new PR, feel free to merge after comment

PR #61, opened 2026-04-15
fix Update Endpoint

KhoaNguyen706:
Just fix Update endpoint

### DIFF
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -155,6 +155,11 @@ func UpdateEventByID(c *gin.Context) {
 		return
 	}
 
+	// Support legacy events that were created before status/visibility defaults.
+	legacyMissingStatus := strings.TrimSpace(existingEvent.Status) == ""
+	legacyMissingVisibility := strings.TrimSpace(existingEvent.Visibility) == ""
+	existingEvent.ApplyDefaults()
+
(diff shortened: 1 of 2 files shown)


pkg/handlers/event.go
```diff
@@ -155,6 +155,11 @@ func UpdateEventByID(c *gin.Context) {
 		return
 	}
 
+	// Support legacy events that were created before status/visibility defaults.
```
steeevin88, line 158: we don't want this, just delete your old events + make a new one

in the future though we probably do need a way to support legacy events when we add new fields... but there's probably a better way to do this

pkg/handlers/event.go
```diff
@@ -203,6 +208,19 @@ func UpdateEventByID(c *gin.Context) {
 		fields["minimum_visible_role"] = updatedEvent.MinimumVisibleRole
 	}
 
+	if legacyMissingVisibility {
```
steeevin88, line 211: remove these

pkg/handlers/event.go
steeevin88: actually, can you be in change of this? let's close this PR and work on #62

steeevin88 (commented):
let's close this PR and work on #62

PR #63, opened 2026-04-17
Add Kafka registration producer and complete outbox publish flow

### LINKED ISSUE #38: Create Producer Routine

codebyemily:
Resolves #38 

Problem
The registration API already persisted pending registration requests in MongoDB and returned 202 Accepted, but a producer is needed to published those requests to Kafka.
The existing consumer expects small reference messages on the registrations topic (request_id, event_id, user_id) so it can load the document from Mongo and finish capacity / duplicate checks.

Solution
pkg/registration/producer.go: 
Add a Producer using segmentio/kafka-go’s Writer, with PublishRegistration sending JSON that matches KafkaRegistrationMessage.
Writer.Close() so pending work is flushed on shutdown.

POST /events/:id/register: 
After CreatePendingRegistration, call PublishRegistration with the generated request_id. If Kafka publish fails
DeleteRegistrationByID removes the pending row and the handler returns 503; on success, still 202 with event_id and request_id.

internal/config: 
Add KAFKA_WRITER_ASYNC (parsed via getEnvBool) and pass it into NewProducer.
Sync writes keep WriteMessages errors meaningful so rollback works; async is optional with different delivery/error behavior.

cmd/server/main.go: 
Wire the producer from config, inject it into RegisterForEvent, use http.Server with SIGINT/SIGTERM handling: graceful Shutdown → producer.Close() → cancel consumer context → consumer.Close() after the consumer goroutine exits.

pkg/db: 
DeleteRegistrationByID for rollback
Ping Mongo after connect to fail fast when the DB is unreachable.

pkg/registration/consumer.go: 
Close() on the reader for shutdown.

docker-compose.yml: 
depends_on with health (Mongo + Kafka), Mongo healthcheck, stable server env for broker/topic/group.

pkg/registration/producer_test.go: 
Unit tests:
TestNewProducer_respectsAsyncArg guards producer constructor / async config
TestKafkaRegistrationMessage_JSONMatchesConsumerContract guards producer ↔ consumer message contract for the Kafka payload.

Screenshot 2026-04-16 at 9 20 55 AM

**Test:**
1. Create Event: New event_id for an isolated test.
2. POST register: Producer path + Mongo pending + 202 + request_id
3. GET status accepted: Consumer processed the message and updated Mongo
4. 4th line in Kafka: Producer message for this registration matches the API.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -4,6 +4,11 @@ import (
 	"context"
 	"log"
 	"net/http"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+	"time"
 
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
@@ -30,7 +35,7 @@ func main() {
 	if err := db.ConnectRedis(cfg.RedisAddr); err != nil {
 		log.Fatalf("Failed to connect to Redis: %v", err)
 	}
-	defer func(){
+	defer func() {
 		if err := db.DisconnectRedis(); err != nil {
 			log.Printf("Error disconnecting Redis: %v", err)
 		}
@@ -38,6 +43,8 @@ func main() {
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+
+	producer := registration.NewProducer([]string{cfg.KafkaBroker}, cfg.KafkaTopic)
 
 	consumer := registration.NewConsumer(
 		[]string{cfg.KafkaBroker},
@@ -71,7 +78,7 @@ func main() {
 		protected.Use(middleware.RequireAuth(middleware.MembershipStateNonMember, cfg.ClientAPIURL))
 		{
 			protected.POST("/", handlers.CreateEvent)
-			protected.POST("/:id/register", handlers.RegisterForEvent)
+			protected.POST("/:id/register", handlers.RegisterForEvent(producer))
 			protected.DELETE("/:id", handlers.DeleteEventByID)
 			protected.PATCH("/:id", handlers.UpdateEventByID)
 		}
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -4,11 +4,6 @@ import (
 	"context"
 	"log"
 	"net/http"
-	"os"
-	"os/signal"
-	"sync"
-	"syscall"
-	"time"
 
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
@@ -44,7 +39,10 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	producer := registration.NewProducer([]string{cfg.KafkaBroker}, cfg.KafkaTopic)
+	producer := registration.NewProducer(
+		[]string{cfg.KafkaBroker},
+		cfg.KafkaTopic,
+	)
 
 	consumer := registration.NewConsumer(
 		[]string{cfg.KafkaBroker},
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,12 @@ services:
       - "8100:27017"
     volumes:
       - ./data/mongo:/data/db
+    healthcheck:
+      test: ["CMD", "mongosh", "--quiet", "--eval", "db.adminCommand('ping')"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+      start_period: 10s
 
   redis:
     image: redis:alpine
@@ -27,18 +33,21 @@ services:
     ports:
       - "${SERVER_PORT:-8002}:${SERVER_PORT:-8002}"
     depends_on:
-      - mongodb
-      - redis
-      - kafka
+      mongodb:
+        condition: service_healthy
+      redis:
+        condition: service_started
+      kafka:
+        condition: service_started
     environment:
-      MONGO_URI: ${MONGO_URI:-mongodb://mongodb:27017}
-      REDIS_ADDR: ${REDIS_ADDR:-redis:6379}
-      KAFKA_BROKER: ${KAFKA_BROKER:-kafka:29092}
-      KAFKA_TOPIC: ${KAFKA_TOPIC:-registrations}
-      KAFKA_GROUP_ID: ${KAFKA_GROUP_ID:-registration-consumers}
-      SERVER_PORT: ${SERVER_PORT:-8002}
-      CLIENT_URL: ${CLIENT_URL:-http://localhost:3000}
-      CLIENT_API_URL: ${CLIENT_API_URL:-http://host.docker.internal:8080}
+      - MONGO_URI=${MONGO_URI:-mongodb://mongodb:27017}
+      - REDIS_ADDR=${REDIS_ADDR:-redis:6379}
+      - KAFKA_BROKER=${KAFKA_BROKER:-kafka:29092}
+      - KAFKA_TOPIC=${KAFKA_TOPIC:-registrations}
+      - KAFKA_GROUP_ID=${KAFKA_GROUP_ID:-registration-consumers}
+      - SERVER_PORT=${SERVER_PORT:-8002}
+      - CLIENT_URL=${CLIENT_URL:-http://localhost:3000}
+      - CLIENT_API_URL=${CLIENT_API_URL:-http://host.docker.internal:8080}
 
   kafka:
     image: apache/kafka:3.8.0
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,12 +5,7 @@ services:
       - "8100:27017"
     volumes:
       - ./data/mongo:/data/db
-    healthcheck:
-      test: ["CMD", "mongosh", "--quiet", "--eval", "db.adminCommand('ping')"]
-      interval: 5s
-      timeout: 5s
-      retries: 5
-      start_period: 10s
+
   redis:
     image: redis:alpine
     ports:
@@ -32,21 +27,18 @@ services:
     ports:
       - "${SERVER_PORT:-8002}:${SERVER_PORT:-8002}"
     depends_on:
-    mongodb:
-      condition: service_healthy
-    redis:
-      condition: service_started
-    kafka:
-      condition: service_healthy
+      - mongodb
+      - redis
+      - kafka
     environment:
-      - MONGO_URI=${MONGO_URI:-mongodb://mongodb:27017}
-      - REDIS_ADDR=${REDIS_ADDR:-redis:6379}
-      - KAFKA_BROKER=${KAFKA_BROKER:-kafka:29092}
-      - KAFKA_TOPIC=${KAFKA_TOPIC:-registrations}
-      - KAFKA_GROUP_ID=${KAFKA_GROUP_ID:-registration-consumers}
-      - SERVER_PORT=${SERVER_PORT:-8002}
-      - CLIENT_URL=${CLIENT_URL:-http://localhost:3000}
-      - CLIENT_API_URL=${CLIENT_API_URL:-http://host.docker.internal:8080}
+      MONGO_URI: ${MONGO_URI:-mongodb://mongodb:27017}
+      REDIS_ADDR: ${REDIS_ADDR:-redis:6379}
+      KAFKA_BROKER: ${KAFKA_BROKER:-kafka:29092}
+      KAFKA_TOPIC: ${KAFKA_TOPIC:-registrations}
+      KAFKA_GROUP_ID: ${KAFKA_GROUP_ID:-registration-consumers}
+      SERVER_PORT: ${SERVER_PORT:-8002}
+      CLIENT_URL: ${CLIENT_URL:-http://localhost:3000}
+      CLIENT_API_URL: ${CLIENT_API_URL:-http://host.docker.internal:8080}
 
   kafka:
     image: apache/kafka:3.8.0
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,31 +1,28 @@
 package config
 
-import (
-	"os"
-	"strings"
-)
+import "os"
 
 type AppConfig struct {
-	MongoURI         string
-	RedisAddr        string
-	KafkaBroker      string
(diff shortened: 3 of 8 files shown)


docker-compose.yml
```diff
@@ -27,16 +33,20 @@ services:
...
-      - KAFKA_GROUP_ID=${KAFKA_GROUP_ID:-registration-consumers}
-      - SERVER_PORT=${SERVER_PORT:-8002}
-      - CLIENT_URL=${CLIENT_URL:-http://localhost:3000}
+      MONGO_URI: mongodb://mongodb:27017
+      REDIS_ADDR: redis:6379
+      KAFKA_BROKER: kafka:29092
+      KAFKA_TOPIC: registrations
+      KAFKA_GROUP_ID: registration-consumers
+      SERVER_PORT: ${SERVER_PORT:-8002}
+      CLIENT_URL: ${CLIENT_URL:-http://localhost:3000}
```
steeevin88: why did we change these lines? we should keep the old implementation unless it was really needed

docker-compose.yml
```diff
@@ -27,16 +33,20 @@ services:
     ports:
       - "${SERVER_PORT:-8002}:${SERVER_PORT:-8002}"
     depends_on:
-      - mongodb
```
steeevin88: could we move the health checks to a separate PR? let's keep this PR solely for producer logic

docker-compose.yml
```diff
@@ -5,20 +5,26 @@ services:
...
     volumes:
       - ./data/redis:/data
     healthcheck:
-      test: ["CMD", "redis-cli", "ping"]
+      test: [ "CMD", "redis-cli", "ping" ]
       interval: 10s
       timeout: 5s
       retries: 3
-      start_period: 5s 
+      start_period: 5s
```
steeevin88, line 27: let's remove this whitespace change

docker-compose.yml
```diff
@@ -5,20 +5,26 @@ services:
...
     image: redis:alpine
     ports:
       - "8101:6379"
-    command: ["redis-server", "--appendonly", "yes"]
+    command: [ "redis-server", "--appendonly", "yes" ]
     volumes:
       - ./data/redis:/data
     healthcheck:
-      test: ["CMD", "redis-cli", "ping"]
+      test: [ "CMD", "redis-cli", "ping" ]
```
steeevin88: let's remove this whitespace change

docker-compose.yml
```diff
@@ -5,20 +5,26 @@ services:
...
+      timeout: 5s
+      retries: 5
+      start_period: 10s
 
   redis:
     image: redis:alpine
     ports:
       - "8101:6379"
-    command: ["redis-server", "--appendonly", "yes"]
+    command: [ "redis-server", "--appendonly", "yes" ]
```
steeevin88: let's remove this whitespace change

docker-compose.yml
```diff
@@ -5,20 +5,26 @@ services:
       - "8100:27017"
     volumes:
       - ./data/mongo:/data/db
+    healthcheck:
```
steeevin88, line 8: as per other comment let's move this healthcheck to another PR

we can merge that PR first

docker-compose.yml
```diff
@@ -58,7 +68,7 @@ services:
       - KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS=0
       - KAFKA_NUM_PARTITIONS=3
     healthcheck:
-      test: ["CMD-SHELL", "/opt/kafka/bin/kafka-broker-api-versions.sh --bootstrap-server localhost:9092 > /dev/null 2>&1"]
+      test: [ "CMD-SHELL", "/opt/kafka/bin/kafka-broker-api-versions.sh --bootstrap-server localhost:9092 > /dev/null 2>&1" ]
```
steeevin88: let's remove this whitespace change

docker-compose.yml
```diff
@@ -75,4 +85,4 @@ services:
       - -c
       - |
         /opt/kafka/bin/kafka-topics.sh --create --topic registrations --bootstrap-server kafka:29092 --partitions 1 --replication-factor 1 --if-not-exists
-        echo "Topic registrations created or already exists"
\ No newline at end of file
+        echo "Topic registrations created or already exists"
```
steeevin88, line 89: agh we probably should have this new line at the end... we can keep this change

pkg/handlers/event.go
```diff
@@ -98,6 +100,7 @@ func CreateEvent(c *gin.Context) {
 
 	createdEvent, err := db.CreateEvent(event)
 	if err != nil {
+		log.Printf("CreateEvent: %v", err)
```
steeevin88: let's be more verbose, ex. `CreateEvent error:`

internal/config/config.go
```diff
@@ -1,26 +1,31 @@
...
-		KafkaTopic:   getEnv("KAFKA_TOPIC", "registrations"),
-		KafkaGroupID: getEnv("KAFKA_GROUP_ID", "registration-consumers"),
-		ServerPort:   getEnv("SERVER_PORT", "8002"),
-		ClientURL:    getEnv("CLIENT_URL", "http://localhost:3000"),
+		MongoURI:         getEnv("MONGO_URI", "mongodb://localhost:27017"),
+		RedisAddr:        getEnv("REDIS_ADDR", "localhost:6379"),
+		KafkaBroker:      getEnv("KAFKA_BROKER", "localhost:9092"),
+		KafkaTopic:       getEnv("KAFKA_TOPIC", "registrations"),
+		KafkaGroupID:     getEnv("KAFKA_GROUP_ID", "registration-consumers"),
+		KafkaWriterAsync: getEnvBool("KAFKA_WRITER_ASYNC", false),
```
steeevin88: i don't like this config

1. we probably will only ever want async or synchronous; never both
2. we probably should prefer synchronous for our use case of registrations. we need to know if Kafka was able to enqueue the registration, such that we're able to send the registrant a response on the frontend (Clark)

let's just remove this config change + default to sync (`Async: false`)

internal/config/config.go
```diff
@@ -30,4 +35,19 @@ func getEnv(key string, fallback string) string {
 		return fallback
 	}
 	return value
-}
\ No newline at end of file
+}
+
+func getEnvBool(key string, fallback bool) bool {
```
steeevin88: let's remove this since we no longer need the new config

cmd/server/main.go
```diff
@@ -38,13 +43,20 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	producer := registration.NewProducer([]string{cfg.KafkaBroker}, cfg.KafkaTopic, cfg.KafkaWriterAsync)
```
steeevin88: nit: can we format this line like we did for the consumer

```
producer := registration.NewProducer(
	[]string{cfg.KafkaBroker}, 
	cfg.KafkaTopic, 
	cfg.KafkaWriterAsync
)
```

steeevin88: `cfg.KafkaWriterAsync` should be removed

pkg/db/mongo.go
```diff
@@ -35,6 +35,13 @@ func Connect(uri string) error {
 		return err
 	}
 
+	pingCtx, pingCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
steeevin88, line 38: similar to other comments, let's move this healthcheck function to a separate PR

pkg/db/registration.go
```diff
@@ -35,6 +35,17 @@ func CreatePendingRegistration(r models.RegistrationRequest) (*models.Registrati
 	return &r, nil
 }
 
+// DeleteRegistrationByID removes a registration document by request_id (used to roll back after a failed outbox publish).
```
steeevin88: we probably don't want this; if a outbox publish failed... we don't actually care that the registration exists in Mongo, we can just leave it

for instnace, it could be helpful in recovery; if a user filled out the form + sent a request, but we failed to process it... then the registrant needs to fill out the form again. in the future... we could use the registration request to "repopulate" the form

pkg/handlers/event.go
```diff
@@ -255,96 +258,106 @@ func UpdateEventByID(c *gin.Context) {
...
+		created, err := db.CreatePendingRegistration(req)
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{
+				"error": "failed to create registration request",
+			})
+			return
+		}
+
+		if err := producer.PublishRegistration(c.Request.Context(), created.RequestID, eventID, payload.Registrant.UserID); err != nil {
+			_ = db.DeleteRegistrationByID(created.RequestID)
```
steeevin88: similar to other comment, let's remove this; we shouldn't delete the registration if enqueuing to Kafka failed

cmd/server/main.go
```diff
@@ -65,13 +77,44 @@ func main() {
...
+
+	if err := producer.Close(); err != nil {
+		log.Printf("Error closing Kafka producer: %v", err)
+	}
+
+	cancel()
+	wg.Wait()
+
+	if err := consumer.Close(); err != nil {
+		log.Printf("Error closing Kafka consumer: %v", err)
```
steeevin88: thanks for being proactive + making our app shutdown more graceful 😄 

a few points though:
1. can we make this a helper function? ex. `gracefulShutdown()`. this function has a lot of logic that is a bit confusing to follow imo
2. could we maybe add a few simple comments just to understand what's going on? i had copilot explain for me 😅 
3. should this maybe be a separate PR instead? i feel like this is separate logic from the intent of this PR (that is, adding the producer). if so, let's make a new PR adding this

cmd/server/main.go
```diff
@@ -38,13 +43,20 @@ func main() {
...
 		cfg.KafkaGroupID,
 	)
 
-	go consumer.Run(ctx)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		consumer.Run(ctx)
+	}()
```
steeevin88: similar to the [comment i made later in this file](https://github.com/SCE-Development/SCEvents/pull/63/changes#r3097677260), if we think the graceful shutdown changes should be in a separate PR, let's move this change as well

pkg/registration/producer_test.go
```diff
@@ -0,0 +1,51 @@
+package registration
```
steeevin88, line 1: thanks for writing tests 😄

steeevin88 (changes requested):
left a few comments, a lot of nits but some key points:

1. i think it's worth considering extracting the "graceful shutdown" logic into a separate PR. while it's related/helpful to the producer logic, it's probably not directly related + should be added on it's own
2. we shouldn't delete registrations if enqueuing a registration request to Kafka (via the producer) fails
3. we should extract the new healthcheck logic in our docker compose (and the ping function) to a new PR (since it's not directly related to the producer logic)
4. let's not have Kafka async, we probably want responses from Kafka (as to if a registration request was enqueued properly or not

feel free to challenge any of these design decisions, just my thoughts
lmk if you want me to clarify anything, regardless thanks for this pr guys 🥇  once the producer is merged we're super close to an MVP

internal/config/config.go
```diff
@@ -32,4 +34,4 @@ func getEnv(key string, fallback string) string {
 		return fallback
 	}
 	return value
-}
+}
```
steeevin88, line 35: i think this entire file is cherry picked? since it's the basically the entire logic of this file let's not cherry pick it, remove the chagnes from this PR

docker-compose.yml
```diff
@@ -5,12 +5,7 @@ services:
       - "8100:27017"
     volumes:
       - ./data/mongo:/data/db
-    healthcheck:
```
steeevin88, line 8: this shouldn't be removed

docker-compose.yml
```diff
@@ -32,21 +27,18 @@ services:
     ports:
       - "${SERVER_PORT:-8002}:${SERVER_PORT:-8002}"
     depends_on:
-    mongodb:
```
steeevin88, line 35: this shouldn't be changed, probably messed up rebase

docker-compose.yml
```diff
@@ -32,21 +27,18 @@ services:
...
-      condition: service_healthy
-    redis:
-      condition: service_started
-    kafka:
-      condition: service_healthy
+      - mongodb
+      - redis
+      - kafka
     environment:
-      - MONGO_URI=${MONGO_URI:-mongodb://mongodb:27017}
```
steeevin88, line 42: this also shouldn't be changed

steeevin88 (changes requested):
almost done, thanks for the update

cmd/server/main.go
```diff
@@ -39,6 +39,10 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	producer := registration.NewProducer(
+		[]string{cfg.KafkaBroker},
+		cfg.KafkaTopic)
```
steeevin88:
```suggestion
	producer := registration.NewProducer(
		[]string{cfg.KafkaBroker},
		cfg.KafkaTopic
	)
```

sorry being really nitpicky 😅

steeevin88 (approved):
one small nit but lgtm

PR #64, opened 2026-04-17
 Add Auto Migration Default

### LINKED ISSUE #62: Create script to update all MongoDB entries when structs change
#61 caught an issue where events created before an Event struct change (ex. #36) would break certain parts of the application, ex. updating an event

while a possible solution is to handle each missing fields manually (ex. as done in #61), this can be really messy if we have a BUNCH of missing fields. and, more importantly, we aren't actually updating the entities in MongoDB! 

the best solution is to have a script that basically says "for a given type (ex. event), if it's missing a certain field, add the DEFAULT VALUE to that field". and it would do it for ALL events

let's introduce a new Go Script that looks through ALL our entities and updates them if they have missing fields:

```
package main

import (
	"context"
	"log"
	"your_project/pkg/database"
	"your_project/pkg/models"
)

func main() {
	// ... database connection setup (db) ...

	// 1. Define the registry of all collections and their corresponding models
	registry := []struct {
		Collection string
		Model      interface{}
	}{
		{"events", models.Event{}},
		{"users", models.User{}},          // Future model
		{"organizations", models.Org{}},   // Future model
		// Just add new models here as the app grows!
	}

	// 2. Loop through and migrate all of them
	for _, reg := range registry {
		col := db.Collection(reg.Collection)
		err := database.AutoMigrateDefaults(context.Background(), col, reg.Model)
		if err != nil {
			log.Fatalf("Failed migrating %s: %v", reg.Collection, err)
		}
	}
	
	log.Println("All defaults migrated successfully!")
}
```

i gpted this but maybe these tasks will help

- [ ] Update Models: Add custom default tags to the fields in pkg/models/event.go (and any other models) that need backfilling (e.g., Status string \bson:"status" json:"status" default:"draft"``).

- [ ] Create Auto-Migrator Utility: Create a new Go function (e.g., AutoMigrateDefaults in pkg/database/migrate.go) that uses reflection to read a struct's bson and default tags, and constructs a MongoDB UpdateMany query for any missing fields.

- [ ] Create Migration Registry Entrypoint: Create a standalone command (e.g., cmd/migrate/main.go) that contains a registry mapping MongoDB collection names to their Go structs (e.g., "events" -> models.Event{}). Iterate through this registry and run the migrator utility for each one.

- [ ] Clean Up Handlers: Once the script is merged and run against the database, remove the legacy lazy-migration checks (like those in PR #61) from the UpdateEventByID handler.

KhoaNguyen706:
Add auto migration

closes  [62](https://github.com/SCE-Development/SCEvents/issues/62)

### DIFF
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -1,0 +1,13 @@
+# cmd
+
+Standalone Go programs. Run with `go run ./cmd/<name>` from the module root.
+
+## migrate
+
+**Purpose:** Backfills MongoDB documents with default values from `default:"..."` struct tags on registered models. Re-running is safe (`$exists: false` filters).
+
+**Usage:**
+
+```bash
+MONGO_URI=mongodb://localhost:27017 go run ./cmd/migrate
+```
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -4,10 +4,19 @@ Standalone Go programs. Run with `go run ./cmd/<name>` from the module root.
 
 ## migrate
 
-**Purpose:** Backfills MongoDB documents with default values from `default:"..."` struct tags on registered models. Re-running is safe (`$exists: false` filters).
+**Purpose**
 
-**Usage:**
+- Backfills MongoDB documents with default values from `default:"..."` struct tags on registered models.
+- Re-running is safe (`$exists: false` filters).
+
+**Usage**
 
 ```bash
-MONGO_URI=mongodb://localhost:27017 go run ./cmd/migrate
+# Set MONGO_URI if you haven't already. By default the server reads it from docker-compose.
+# Example when Mongo is published on the host (see ports in docker-compose.yml):
+# MONGO_URI=mongodb://localhost:8100
+
+export MONGO_URI=mongodb://localhost:8100
+
+go run ./cmd/migrate
 ```
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -11,12 +11,27 @@ Standalone Go programs. Run with `go run ./cmd/<name>` from the module root.
 
 **Usage**
 
+The first argument is required: which collection to migrate (`events` or `registrations`). Omitting it prints `Please specify a collection` and exits; an unknown name prints `Provided type doesn't exist`.
+
 ```bash
 # Set MONGO_URI if you haven't already. By default the server reads it from docker-compose.
-# Example when Mongo is published on the host (see ports in docker-compose.yml):
-# MONGO_URI=mongodb://localhost:8100
+# Example when Mongo is published on the host (docker-compose maps 8100 -> 27017):
+# export MONGO_URI=mongodb://localhost:8100
 
 export MONGO_URI=mongodb://localhost:8100
 
-go run ./cmd/migrate
+go run ./cmd/migrate events
+go run ./cmd/migrate registrations
 ```
+
+**Examples**
+
+- `events` — runs defaults migration against the `events` collection using `models.Event` (e.g. backfill `admins`, `registration_form`, etc. when those fields were added with `default` tags).
+- `registrations` — runs the same logic against the `registrations` collection using `models.RegistrationRequest`. It only updates fields that have a `default` struct tag; if none exist yet, the run succeeds with zero backfills.
+
+**Testing locally**
+
+1. Start Mongo (e.g. `docker compose up -d mongodb`) and set `MONGO_URI` to match your host port (often `mongodb://localhost:8100` from this repo’s compose file).
+2. Seed or reuse documents that are missing a field you recently added with a `default` tag (older rows won’t have that key in BSON).
+3. Run `go run ./cmd/migrate events` (or `registrations`) and check the log for per-field backfill counts.
+4. Optional: in `mongosh`, open the collection and confirm missing keys were set (e.g. `db.events.findOne()`).
--- a/cmd/migrate/main.go
+++ b/cmd/migrate/main.go
@@ -12,17 +12,7 @@ import (
 	"github.com/SCE-Development/SCEvents/pkg/models"
 )
 
-type registryEntry struct {
-	Collection string
-	Model      any
-}
 
-// registry maps CLI subcommands to a Mongo collection and the struct used for default tags.
-// Add an entry when a new collection gets fields with default tags; Go cannot infer model types from a string alone.
-var registry = map[string]registryEntry{
-	"events":        {Collection: "events", Model: models.Event{}},
-	"registrations": {Collection: "registrations", Model: models.RegistrationRequest{}},
-}
 
 func main() {
 	if len(os.Args) < 2 || strings.TrimSpace(os.Args[1]) == "" {
@@ -31,7 +21,7 @@ func main() {
 	}
 
 	key := strings.ToLower(strings.TrimSpace(os.Args[1]))
-	entry, ok := registry[key]
+	entry, ok := models.MigrationRegistry[key]
 	if !ok {
 		log.Print("Provided type doesn't exist")
 		os.Exit(1)
--- a/cmd/migrate/main.go
+++ b/cmd/migrate/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"log"
+	"os"
+	"strings"
 
 	"github.com/SCE-Development/SCEvents/internal/config"
 	"github.com/SCE-Development/SCEvents/pkg/database"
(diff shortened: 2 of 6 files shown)


pkg/database/migrate.go
```diff
@@ -0,0 +1,131 @@
...
+		return n, nil
+
+	case reflect.Bool:
+		b, err := strconv.ParseBool(raw)
+		if err != nil {
+			return nil, fmt.Errorf("invalid bool default %q: %w", raw, err)
+		}
+		return b, nil
+
+	case reflect.Slice:
```
steeevin88, line 99: according to gemini this doesn't support lists if they have default values?

ex. if a field had this: `default:"[\"legacy\", \"imported\"]"`, this code wouldn't work i believe

maybe something like this?
```
// Add "encoding/json" to the imports at the top of the file

case reflect.Slice:
	// Create a pointer to a new slice of the target type (e.g., *[]string)
	slicePtr := reflect.New(t)
	
	// Try to parse the default string as JSON (e.g., `["admin", "member"]` or `[]`)
	if err := json.Unmarshal([]byte(raw), slicePtr.Interface()); err != nil {
		return nil, fmt.Errorf("failed to parse slice default %q as JSON: %w", raw, err)
	}
	
	// Dereference the pointer and return the actual slice
	return slicePtr.Elem().Interface(), nil
```

pkg/models/event.go
```diff
@@ -47,12 +47,12 @@ type Event struct {
 	Time               string         `bson:"time" json:"time"`
 	Location           string         `bson:"location" json:"location"`
 	Description        string         `bson:"description" json:"description"`
-	Admins             []string       `bson:"admins" json:"admins"`
-	RegistrationForm   []FormQuestion `bson:"registration_form" json:"registration_form"`
-	MaxAttendees       int            `bson:"max_attendees" json:"max_attendees"`
+	Admins             []string       `bson:"admins" json:"admins" default:"[]"`
```
steeevin88, line 51: hmm... we can make default an empty array for now. but we should probably... have a way that supports ANY admin access? because if we set the default to an empty list of admins, it means that NOBODY can edit the event.

obviously we could manually edit the MongoDB entry to add an admin but this is tedious

i'll open an issue for this: #65

but yeah we can leave this as the default

cmd/migrate/main.go
```diff
@@ -0,0 +1,62 @@
+// Command migrate backfills MongoDB documents with default values declared
+// via `default:"..."` struct tags on registered models.
+//
+// Usage:
+//
+//	MONGO_URI=mongodb://localhost:27017 go run ./cmd/migrate
+//
+// Re-running is safe: UpdateMany uses {$exists: false}, so already-migrated
+// documents are not touched.
```
steeevin88: can we move this to a README.md in the `/cmd` dir? let's have a dedicated instructions file

let's format it where there is a header for each script that we may have in the `/cmd` directory... right now there will only be one

steeevin88: let's keep the readme concise though... maybe for each script we have two sections:
- Purpose --> one sentence summary of the script's purpose
- Usage --> instructions on using the script

pkg/database/migrate.go
```diff
@@ -0,0 +1,131 @@
+// Package database hosts cross-cutting MongoDB helpers that aren't tied to
+// a specific collection, starting with the auto-migration utility that
+// backfills documents using `default:"..."` struct tags on models.
```
steeevin88: let's remove these comments

pkg/database/migrate.go
```diff
@@ -0,0 +1,131 @@
...
+//	coll.UpdateMany({field: {$exists: false}}, {$set: {field: <parsedDefault>}})
+//
+// Fields without a `default` tag (or without a usable `bson` name) are
+// skipped, so required columns like "_id" / "name" are never backfilled.
+//
+// Supported field kinds:
+//   - string: raw tag value
+//   - intN / uintN: parsed via strconv
+//   - bool: parsed via strconv.ParseBool
+//   - slice: only "[]" is accepted, written as an empty typed slice
```
steeevin88: let's remove the "supported fields" comment, we can infer this from the code

steeevin88 (changes requested):
awesome stuff! besides the comments, a few more points:

1. can you change the last sentence in your PR description? it should be "closes", not "closed". closed implies that we close the issue before merging this PR. closes implies that we WILL close the issue after merging this PR
2. did you test this locally? i know it's a bit weird to test but since you already had data that wasn't up-to-date i'm guessing you ran the script on it and it worked? just want to confirm quickly

cmd/README.md
```diff
@@ -0,0 +1,13 @@
+# cmd
+
+Standalone Go programs. Run with `go run ./cmd/<name>` from the module root.
+
+## migrate
+
+**Purpose:** Backfills MongoDB documents with default values from `default:"..."` struct tags on registered models. Re-running is safe (`$exists: false` filters).
```
steeevin88: i'm just being nitpicky but i'd prefer if this section was bullet points, instead. can we do:

```
**Purpose** 
- Backfills MongoDB documents with default values from `default:"..."` struct tags on registered models
- Re-running is safe (`$exists: false` filters).
```

cmd/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+**Usage:**
+
+```bash
+MONGO_URI=mongodb://localhost:27017 go run ./cmd/migrate
```
steeevin88: let's make each command a seprate line, and maybe have a comment above the .env. something like:

```bash
# set MONGO_URI if you haven't already. by default, this value is declared in the docker-compose
# MONGO_URI=mongodb://localhost:27017

go run ./cmd/migrate
```

steeevin88 (changes requested):
cool, and again was this tested? just wanna make sure

steeevin88:
1. main got a new commit, can you rebase?
2. can you also capitalize your PR title

pkg/models/event.go
```diff
@@ -47,12 +47,13 @@ type Event struct {
 	Time               string         `bson:"time" json:"time"`
 	Location           string         `bson:"location" json:"location"`
 	Description        string         `bson:"description" json:"description"`
-	Admins             []string       `bson:"admins" json:"admins"`
-	RegistrationForm   []FormQuestion `bson:"registration_form" json:"registration_form"`
-	MaxAttendees       int            `bson:"max_attendees" json:"max_attendees"`
+	// Admins lists event-specific editors. Empty slice after migration means no dedicated admins yet; see #65 for planned "site admin may edit" behavior.
```
steeevin88: can you not have the `see #65 for planned...` comment lol
and let's place the comment after line 51 instead?

```
Admins             []string       `bson:"admins" json:"admins" default:"[]"` // empty admin list means any admin can edit
```

cmd/migrate/main.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/SCE-Development/SCEvents/pkg/models"
+)
+
+type registryEntry struct {
+	Collection string
+	Model      any
+}
+
+var registry = []registryEntry{
+	{Collection: "events", Model: models.Event{}},
```
steeevin88: wait so does this only support Events right now?

and does this script update ALL collections that we've specified? i'd prefer to add an argument that let's us specify which struct we are updating...

ex. `go run ./cmd/migrate events` updates the events collection, whereas `go run ./cmd/migrate registrations` updates the registrations collection

so in other words:
- we should allow the other structs that we have in mongo --> registrations 
- add a way to specify what type. if no type is specified, we should jsut log a message ("Please specify a collection"). and if an invalid type is specified (one that doesn't exist) we should log another message ("Provided type doesn't exist")

pkg/models/event.go
steeevin88: on second thought let's just remove this comment and let's add it in your other PR: #67

cmd/README.md
```diff
@@ -0,0 +1,22 @@
...
+**Usage**
+
+```bash
+# Set MONGO_URI if you haven't already. By default the server reads it from docker-compose.
+# Example when Mongo is published on the host (see ports in docker-compose.yml):
+# MONGO_URI=mongodb://localhost:8100
+
+export MONGO_URI=mongodb://localhost:8100
+
+go run ./cmd/migrate
```
steeevin88: shouldn't this be updated now? since there are required args?

steeevin88: can you give examples of both usecases, so events + registrations

cmd/migrate/main.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/SCE-Development/SCEvents/pkg/db"
+	"github.com/SCE-Development/SCEvents/pkg/models"
+)
+
+type registryEntry struct {
+	Collection string
+	Model      any
+}
+
+var registry = map[string]registryEntry{
```
steeevin88: is there a way to link our models better? instead of having to hardcode the types, it'd be useful for when we add new structs

if it's not possible though then we can just leave it

steeevin88 (changes requested):

steeevin88:
also remember to update the PR title lol let's capitalize "Add"

steeevin88 (approved):
i think this looks fine, if it gets used we can debug in the future

Issue #41, opened 2026-04-12
Determine who can update events


steeevin88:
current to-be-added logic will only allow admins of an event to update/edit an event; however, this matches weird with Clark's membership levels:

https://github.com/SCE-Development/Clark/blob/ea70088c8baee4cd2caa8c5d6d0ddfa3f1cd481b/src/Enums.js#L23-L30

i think the ideal flow would be:
- if user is an admin, they can edit ANY events
- if user is an officer, they can edit event's they are admins of
- anything lower --> cannot edit events

but... i'm not sure how to do this since the membership levels live in Clark's MongoDB instance. there probably is a way but we'll need to do some research, and this honestly isn't a big priority

**please reply with your gameplan in a comment to this / DM me**

KhoaNguyen706:
So Do I need to make a UI page in Clark to hanlde update events ?

KhoaNguyen706:
The Gameplan (Auth V2)
Since Nginx sits as the reverse proxy gateway between the Clark frontend and our microservices, we can use it to perfectly bridge the gap between Clark's MongoDB membership levels and SCEvents without implementing JWT logic in the events server!

Nginx Auth Request Integration: Nginx currently coordinates traffic. We can configure Nginx to use the auth_request module. When a request comes to /api/events/*, Nginx will first securely bounce the user's session cookie/JWT to a lightweight internal Clark Auth token verifier.

Injecting Upstream Headers: When Clark verifies the token, it returns a 200 OK to Nginx and passes the user’s MongoDB context in the headers (e.g., X-User-Id, X-Access-Level). Nginx will then proxy the original request to the SCEvents backend, automatically securely injecting those headers!

SCEvents Minimal Backend Updates: Because Nginx guarantees those headers exist and are verified, the 

UpdateEvent
 handler in 

handlers/event.go
 becomes incredibly simple and lightweight:

Read c.GetHeader("X-Access-Level") and c.GetHeader("X-User-Id").
If X-Access-Level == 3 (Admin): Bypass checks, allow the edit immediately.
If X-Access-Level == 2 (Officer): Verify X-User-Id is present inside the event's event.Admins array. If yes, allow edit.
Else: Reject with 403 Forbidden.
Frontend Alignment: On the Clark frontend, we simply check 

useSCE()
:

Admin (3): Display the Pencil icon on every Event Card.
Officer (2): Display the Pencil icon only if user._id matches event.admins.
[...]

PR #67, opened 2026-04-18
Support admin-less events 

### LINKED ISSUE #65: Support "admin-less' events
#64 adds a default to the `Event` struct's list of admins; that is, an empty list

that being said... this means nobody can edit the event. we don't want that... the solution is to manually edit the event entry in the MongoDB container... but that totally sucks

we should have a solution that
- sets the default value of an event... as "admin-less", meaning any ADMIN can modify the event
- then... if someone adds a single admin to the event, only that admin can edit the event

i think an easy solution would just be to say `if Event.admins == []`, then all admins can edit the event. otherwise, only the admins in `Event.admins`... we could do this? since realistically we should never have an event with no admins... my only gripe is that, logically, this is a bit confusing
- no admins == unlimited admins? i guess it kind of makes sense? let's make sure to leave a comment on the event struct then

KhoaNguyen706:
- Event.CanEdit: when Admins is empty, any site admin (user_role=admin) can edit; otherwise only listed admins.
- Event.IsListedAdmin: renamed from IsAdmin for clarity.
- PATCH/DELETE handlers: read user_id and user_role from query, return 401 when unauthenticated and 403 when unauthorized via CanEdit.
- Added doc comment on Admins field pointing at issue #65.
Admin Less tesing:
image
No auth:
image
Token failed:
image

Closes #65

### DIFF
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -18,6 +18,18 @@ import (
 
 const dateLayout = "2006-01-02"
 
+func writeEventEditForbidden(c *gin.Context, ev *models.Event) {
+	if len(ev.Admins) == 0 {
+		c.JSON(http.StatusForbidden, gin.H{
+			"error": "this event has no dedicated admins; only site admins may modify it",
+		})
+		return
+	}
+	c.JSON(http.StatusForbidden, gin.H{
+		"error": "you are not an admin of this event",
+	})
+}
+
 // GetEvents: query startDate & endDate (YYYY-MM-DD), or omit both for current UTC month; one alone is 400.
 func GetEvents(c *gin.Context) {
 	startQ := strings.TrimSpace(c.Query("startDate"))
@@ -121,7 +133,23 @@ func CreateEvent(c *gin.Context) {
 func DeleteEventByID(c *gin.Context) {
 	id := c.Param("id")
 
-	err := db.DeleteEventByID(id)
+	userID := c.GetString("userID")
+	userRole := c.GetString("userRole")
+
+	existingEvent, err := db.GetEventByID(id)
+	if err != nil {
+		c.JSON(http.StatusNotFound, gin.H{
+			"error": "event not found",
+		})
+		return
+	}
+
+	if !existingEvent.CanEdit(userID, userRole) {
+		writeEventEditForbidden(c, existingEvent)
+		return
+	}
+
+	err = db.DeleteEventByID(id)
 	if err != nil {
 		c.JSON(http.StatusNotFound, gin.H{
 			"error": "event not found",
@@ -138,18 +166,9 @@ func DeleteEventByID(c *gin.Context) {
 func UpdateEventByID(c *gin.Context) {
 	id := c.Param("id")
 
-	// Basic event-admin auth check:
-	// The caller must pass their user_id as a query parameter.
-	// We verify they are listed in the event's Admins before allowing the update.
-	userID := c.Query("user_id")
-	if userID == "" {
-		c.JSON(http.StatusUnauthorized, gin.H{
-			"error": "missing user_id query parameter",
-		})
-		return
-	}
+	userID := c.GetString("userID")
+	userRole := c.GetString("userRole")
 
-	// Fetch the existing event to verify admin access
 	existingEvent, err := db.GetEventByID(id)
 	if err != nil {
 		c.JSON(http.StatusNotFound, gin.H{
@@ -158,11 +177,8 @@ func UpdateEventByID(c *gin.Context) {
 		return
 	}
 
-	// Check if user is an admin of this event
-	if !existingEvent.IsAdmin(userID) {
-		c.JSON(http.StatusForbidden, gin.H{
-			"error": "you are not an admin of this event",
-		})
+	if !existingEvent.CanEdit(userID, userRole) {
+		writeEventEditForbidden(c, existingEvent)
 		return
 	}
 
--- a/pkg/models/event.go
+++ b/pkg/models/event.go
@@ -192,6 +192,15 @@ func (e *Event) IsAdmin(userID string) bool {
 	return false
 }
 
+// CanEdit reports whether the caller may modify this event.
+// Admin-less events require callerSiteRole=admin; otherwise userID must be in Admins.
+func (e *Event) CanEdit(userID, callerSiteRole string) bool {
+	if len(e.Admins) == 0 {
+		return strings.EqualFold(strings.TrimSpace(callerSiteRole), RoleAdmin)
+	}
+	return e.IsAdmin(userID)
+}
+
 func (e *Event) ValidateRegistration(answers map[string]any) error {
 	for _, question := range e.RegistrationForm {
 		if !question.Required {


pkg/models/event.go
```diff
@@ -47,8 +47,13 @@ type Event struct {
...
 	Description        string         `bson:"description" json:"description"`
-	// Admins lists event-specific editors. Empty slice after migration means no dedicated admins yet; see #65 for planned "site admin may edit" behavior.
-	Admins             []string       `bson:"admins" json:"admins" default:"[]"`
+	// Admins lists user IDs allowed to edit this event.
+	//
+	// Admin-less (empty or nil slice): any site admin may edit; the API expects callers
+	// trusted by the gateway (e.g. Clark) to pass user_role=admin on PATCH/DELETE.
+	// Once at least one ID is listed here, only those users may edit (site admin is not
+	// implicitly included—add yourself to Admins if you need access).
+	Admins []string `bson:"admins" json:"admins" default:"[]"`
```
steeevin88: this comment block is way too long. let's shorten it

because it'll be multi-line we can keep it above line 56

but let's reduce it down to two lines please, ex.:

```
// empty list of admins = all user IDs with admin perms are admins of the event 
// otherwise (admins isn't empty), only specified user IDs are admins
```

pkg/models/event.go
```diff
@@ -173,16 +178,29 @@ func ValidateMinimumVisibleRole(visibility string, role string) error {
...
 }
 
-// IsAdmin checks if the given userID is in the event's Admins list.
-func (e *Event) IsAdmin(userID string) bool {
-	for _, admin := range e.Admins {
-		if admin == userID {
+// IsListedAdmin reports whether userID appears in this event's Admins list.
+func (e *Event) IsListedAdmin(userID string) bool {
+	userID = strings.TrimSpace(userID)
+	for _, admin := range e.Amins {
```
steeevin88: typo? `e.Amins` instead of `e.Admins`

pkg/models/event.go
```diff
@@ -173,16 +178,29 @@ func ValidateMinimumVisibleRole(visibility string, role string) error {
...
 	return false
 }
 
+// CanEdit reports whether userID may modify this event (PATCH/DELETE).
+//
+// When Admins is empty ("admin-less"), only callers asserting the site admin role may
+// edit; pass callerSiteRole as RoleAdmin ("admin"), e.g. query user_role=admin from Clark.
+// When Admins is non-empty, only listed user IDs may edit.
+func (e *Event) CanEdit(userID, callerSiteRole string) bool {
+	if len(e.Amins) == 0 {
```
steeevin88: typo? e.Amins instead of e.Admins

pkg/models/event.go
```diff
@@ -173,16 +178,29 @@ func ValidateMinimumVisibleRole(visibility string, role string) error {
...
 		}
 	}
 	return false
 }
 
+// CanEdit reports whether userID may modify this event (PATCH/DELETE).
+//
+// When Admins is empty ("admin-less"), only callers asserting the site admin role may
+// edit; pass callerSiteRole as RoleAdmin ("admin"), e.g. query user_role=admin from Clark.
+// When Admins is non-empty, only listed user IDs may edit.
```
steeevin88: again this comment is way too long, let's keep it at max two lines + be more concise please...

pkg/models/event.go
```diff
@@ -173,16 +178,29 @@ func ValidateMinimumVisibleRole(visibility string, role string) error {
...
 
-// IsAdmin checks if the given userID is in the event's Admins list.
-func (e *Event) IsAdmin(userID string) bool {
-	for _, admin := range e.Admins {
-		if admin == userID {
+// IsListedAdmin reports whether userID appears in this event's Admins list.
+func (e *Event) IsListedAdmin(userID string) bool {
+	userID = strings.TrimSpace(userID)
+	for _, admin := range e.Amins {
+		if strings.TrimSpace(admin) == userID {
```
steeevin88: actually, why did we even change this function? i don't think the implementation changed, can't we just use the old function

pkg/handlers/event.go
```diff
@@ -135,9 +169,8 @@ func DeleteEventByID(c *gin.Context) {
 func UpdateEventByID(c *gin.Context) {
 	id := c.Param("id")
 
-	// Basic event-admin auth check:
-	// The caller must pass their user_id as a query parameter.
-	// We verify they are listed in the event's Admins before allowing the update.
+	// Caller must pass user_id. For admin-less events (empty Admins), trusted clients
+	// must also pass user_role=admin so site admins can edit legacy/migrated events.
```
steeevin88: this means we need to update Clark right? if so I'll open a new issue for this

steeevin88: actually... this is a part of #41 lol, i'll update it to make a plan + make corresponding issues in Clark

pkg/handlers/event.go
```diff
@@ -118,7 +130,29 @@ func CreateEvent(c *gin.Context) {
...
+			"error": "event not found",
+		})
+		return
+	}
+
+	userRole := strings.TrimSpace(c.Query("user_role"))
+	if !existingEvent.CanEdit(userID, userRole) {
+		writeEventEditForbidden(c, existingEvent)
+		return
+	}
```
steeevin88:
```suggestion
	isClarkAdmin := nil // TODO: update once Clark auth endpoint is added
	if !existingEvent.CanEdit(userID, isClarkAdmin ) {
		writeEventEditForbidden(c, existingEvent)
		return
	}
```

steeevin88 (commented):
a few comments

i'm fine with merging this now but this definitely won't be our end-logic b/c it's technically not safe 😅 we need to properly verify that the `user-role` isn't just a spoofed value

#41 will solve this but it's a bit complicated... we can just merge this now + fix it

steeevin88:
can you also remoe `#65` from the PR title

pkg/handlers/event.go
```diff
@@ -118,7 +130,29 @@ func CreateEvent(c *gin.Context) {
...
 	id := c.Param("id")
 
-	err := db.DeleteEventByID(id)
+	userID := c.Query("user_id")
+	if userID == "" {
+		c.JSON(http.StatusUnauthorized, gin.H{
+			"error": "missing user_id query parameter",
+		})
+		return
+	}
```
steeevin88: now that #41 was resolved, we should grab the userID straight from context as it's guaranteed to be there as per the middleware added in #69 

same with the `user_role`; let's grab them both here using `getString()` (i think?) and handle it accordingly

pkg/handlers/event.go
```diff
@@ -155,11 +187,9 @@ func UpdateEventByID(c *gin.Context) {
 		return
 	}
 
-	// Check if user is an admin of this event
-	if !existingEvent.IsAdmin(userID) {
-		c.JSON(http.StatusForbidden, gin.H{
-			"error": "you are not an admin of this event",
-		})
+	userRole := strings.TrimSpace(c.Query("user_role"))
```
steeevin88: see previous comment

steeevin88 (commented):
was this tested? if not can we test this + provide screenshots in the PR description 😄 

also this is targeted to merge into your branch for #64 , but we should probably just merge this into main? not sure how to "redirect" it. we can merge either first


edit: to change the branch you can just edit the PR + it'll let you change the target branch

image

steeevin88 (approved):
awesome, thanks for fixing everything

PR #70, opened 2026-04-18
Add health check for MongoDB connection and Docker compose health checks

codebyemily:


steeevin88 (approved):
did you want to add this from the other PR as well?

https://github.com/SCE-Development/SCEvents/blob/328bd548ebd7103c4e351561b4e4f241d710e4d3/pkg/db/mongo.go#L38-L43

if so let's change the PR title

also small nit, could we capitalize the PR title, otherwise lgtm thanks 😄

PR #72, opened 2026-04-19
Add backend support for event waitlisting

### LINKED ISSUE #39: Support event waitlisting
if an event is full... we should allow users to join a waitlist (if the event permits the waitlist)

- [ ] update the event schema to support having a waitlist
- [ ] update the event creation flow to support having a waitlist
  - event creators should be able to not have a waitlist for the event
  - i think this will probably just be an additional boolean field for an event? or something else you can come up with
- [ ] create an endpoint that adds a user to an event's waitlist
  - we can just write directly to MongoDB here...

a few more points:
1. we'll need to open a PR in Clark to render the waitlist logic as well
2. should there be a waitlist size? you can make this decision

miiiyyi:
Closes #39 
What changed 
- added `waitlist_enabled` and `waitlist_size` to the `Event` model
- added validation/default behavior for waitlist fields
- added a new `WaitlistEntry` model
- added MongoDB helpers for creating/checking/counting waitlist entries
- added `POST /events/:id/waitlist`
- wired the new waitlist route into the protected event routes

Behavior
- event creators can choose whether an event supports a waitlist
- if waitlist is enabled, a positive waitlist size is required
- authenticated users can join the waitlist for an event
- users cannot join if:
  - the event does not allow a waitlist
  - the waitlist is full
  - they are already registered
  - they are already on the waitlist

Update:
- added waitlist indexes:
  - unique `(event_id, user_id)`
  - index on `event_id`
- added duplicate-key handling so duplicate waitlist joins return a conflict instead of a server error


pkg/handlers/event.go
```diff
@@ -347,6 +347,106 @@ func RegisterForEvent(c *gin.Context) {
...
+		EventID: eventID,
+		UserID:  userID,
+	}
+
+	if err := db.CreateWaitlistEntry(entry); err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{
+			"error": "failed to join waitlist",
+		})
+		return
+	}
```
steeevin88, line 461: gemini says there is a potential race condition here, is this true? 

if so let's add logic to handle it (ex. mongodb transactions? idk)

steeevin88: or if it's too complicated we can just skip it lol i doubt we'll ever hit this scenario

pkg/db/waitlist.go
```diff
@@ -0,0 +1,61 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	filter := bson.M{
+		"event_id": eventID,
+	}
+
+	return coll.CountDocuments(ctx, filter)
+}
```
steeevin88, line 85: these are potentially super slow queries (ex. if our waitlist collection is super super long)

we should probably index the waitlist collection on the eventId and userID

let's add a new file in `db`, ex. `waitlist.go` that initializes indexes for the waitlist collection

steeevin88 (changes requested):

pkg/handlers/event.go
miiiyyi: Yeah there is still a race window around strict waitlist capacity enforcement for two different users. I added a unique `(event_id, user_id)` index plus duplicate-key handling, so duplicate joins for the same user are handled cleanly now, and I also added the requested indexes for the waitlist queries. A follow-up PR for capacity enforcement / transactions might be better since this could be a bit complicated.

pkg/db/waitlist_indexes.go
```diff
@@ -0,0 +1,33 @@
...
+			Options: options.Index().SetUnique(true),
+		},
+		{
+			Keys: bson.D{
+				{Key: "event_id", Value: 1},
+			},
+		},
+	})
+	return err
+}
```
steeevin88: im slow i didn't realize `waitlist.go` was added in this PR
could we just move this function to the other file to keep the files clean, sorry about that

steeevin88 (approved):
besides moving the function lgtm! feel free to merge after rebasing

PR #74, opened 2026-04-19
Add gracefulshutdown function

### LINKED ISSUE #73: Support graceful shutdowns
as per convo in https://github.com/SCE-Development/SCEvents/pull/63, extract the logic used to gracefully shutdown SCEvents into a different PR

codebyemily:
Resolves #73 

**Problem**
                                                                                                                                                                                
  The server had no graceful shutdown logic. When receiving a termination signal (e.g. docker stop, Kubernetes pod eviction, Ctrl+C), the process would exit immediately —
  dropping in-flight HTTP requests, leaving the Kafka consumer goroutine abandoned with no way to wait for it to finish, and skipping MongoDB/Redis disconnect cleanup.

  There was also a secondary bug: the HTTP server was started with r.Run(), which calls log.Fatalf on failure internally. If the server failed to bind (e.g. port already in
  use), os.Exit would be called directly, bypassing all defers and skipping DB cleanup. The depends_on block in docker-compose.yml was also missing indentation, so the server
  service wasn't actually waiting on MongoDB, Redis, or Kafka to be healthy before starting.

  **Solution**

  cmd/server/main.go
  - Replaced r.Run() with an http.Server started in a goroutine, sending startup errors to a channel instead of calling log.Fatalf, so defers always run on failure
  - Added gracefulShutdown() helper that blocks on either a SIGINT/SIGTERM signal or a server error, then:
    a. Drains in-flight HTTP requests with a 15-second timeout via srv.Shutdown()
    b. Cancels the shared context to stop the Kafka consumer
    c. Waits on a sync.WaitGroup for the consumer goroutine to fully exit before the process returns
  - Wrapped the Kafka consumer goroutine in a WaitGroup so shutdown can block until it exits cleanly

  pkg/registration/consumer.go
  - Added Close() method to Consumer to cleanly close the underlying kafka-go reader after the goroutine exits

  docker-compose.yml
  - Fixed indentation on the server service's depends_on block so it correctly waits for MongoDB, Redis, and Kafka to be healthy before starting

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -4,6 +4,11 @@ import (
 	"context"
 	"log"
 	"net/http"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+	"time"
 
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
@@ -37,7 +42,6 @@ func main() {
 	}()
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
 
 	producer := registration.NewProducer(
 		[]string{cfg.KafkaBroker},
@@ -50,7 +54,12 @@ func main() {
 		cfg.KafkaGroupID,
 	)
 
-	go consumer.Run(ctx)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		consumer.Run(ctx)
+	}()
 
 	r := gin.Default()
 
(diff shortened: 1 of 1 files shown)


cmd/server/main.go
```diff
@@ -77,7 +86,50 @@ func main() {
...
+	if err := srv.Shutdown(shutdownCtx); err != nil {
+		log.Printf("HTTP server shutdown: %v", err)
+	}
+
+	// Stop the Kafka consumer and wait for its goroutine to exit.
+	cancel()
+	wg.Wait()
+
+	if err := consumer.Close(); err != nil {
+		log.Printf("Error closing Kafka consumer: %v", err)
```
steeevin88: i feel like having the function `gracefulShutdown()` start the server is a little bit weird. also, it's a little bit rigid; if we happen to have more goroutines (in addition to the consumer) that need to be shut down along with the server, we'd have to manually add each one

can we maybe do something like this:
```go
...
...
...
	srv := &http.Server{
			Addr:    ":" + cfg.ServerPort,
			Handler: r,
		}

	// Create a channel to catch critical errors from background processes
	errChan := make(chan error, 1)

	// Start the HTTP server in a goroutine
	go func() {
		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
			errChan <- err
		}
	}()

	// Define our cleanup functions
	closeServer := func(ctx context.Context) error {
		return srv.Shutdown(ctx)
	}

	closeConsumer := func(ctx context.Context) error {
		// consumer.Close() doesn't take a context, so we just wrap it
		return consumer.Close()
	}

	// Block and wait for shutdown
	gracefulShutdown(cancel, &wg, errChan, closeServer, closeConsumer)
}


// gracefulShutdown blocks until a stop signal is received or a critical error occurs.
// It then cancels the global context and executes all provided closer functions with a timeout.
func gracefulShutdown(cancel context.CancelFunc, wg *sync.WaitGroup, errChan <-chan error, closers ...func(context.Context) error) {
	quit := make(chan os.Signal, 1)
	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)

	// Block until a signal is received or a background service crashes
	select {
	case err := <-errChan:
		log.Printf("critical error received: %v", err)
	case <-quit:
		log.Println("shutting down gracefully...")
	}

	// 1. Cancel the global context to tell background workers (like Kafka) to stop looping
	cancel()

	// 2. Give cleanup tasks up to 15 seconds to finish
	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 15*time.Second)
	defer shutdownCancel()

	// 3. Execute all provided closer functions
	for _, closer := range closers {
		if err := closer(shutdownCtx); err != nil {
			log.Printf("cleanup error: %v", err)
		}
	}

	// 4. Wait for all goroutines (tracked by wg) to finish
	wg.Wait()
	log.Println("shutdown complete.")
}
```
feel free to rewrite of course, i just grabbed this from gemini lol but let's try to abstract a bit, the key points are:

- `gracefulShutdown` accepts a list of `closers`, so as opposed to very specific things to close, it's just a general list of "closing functions" that it knows it needs to close. this scales better if we add more routines in the future (ex. maybe `gracefulShutdown(cancel, &wg, errChan, closeServer, closeConsumer, closeRedis, closeCron)`
- the server is started in `main.go` but `gracefulShutdown` will shut it down gracefully

pkg/registration/consumer.go
```diff
@@ -49,6 +49,10 @@ func (c *Consumer) Run(ctx context.Context) {
 	}
 }
 
+func (c *Consumer) Close() error {
```
steeevin88, line 53: i think this can be removed once #63 is merged

docker-compose.yml
```diff
@@ -32,12 +32,12 @@ services:
     ports:
       - "${SERVER_PORT:-8002}:${SERVER_PORT:-8002}"
     depends_on:
-    mongodb:
-      condition: service_healthy
-    redis:
-      condition: service_started
-    kafka:
-      condition: service_healthy
+      mongodb:
```
steeevin88, line 36: i think this can be removed once #63 is merged

steeevin88 (changes requested):
hope these make sense! this definitely works but maybe we can alter the implementation to future-proof it more

cmd/server/main.go
```diff
@@ -96,45 +96,52 @@ func main() {
...
 	go func() {
 		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-			serverErr <- err
+			errChan <- err
 		}
 	}()
 
-	// Block until we receive a termination signal or the server errors out.
+	// Block until a signal is received or a background service crashes
+	gracefulShutdown(cancel, &wg, errChan,
```
codebyemily, line 109: I think since srv.Shutdown is only used once directly into gracefulShutdown, there's no need to create a variable. Also, srv.Shutdown matches func(context.Context) error directly, so it's passed as-is with no wrapper needed.

steeevin88 (approved):
awesome, thanks!

Issue #54, opened 2026-04-15
Support mocking external services for unit tests


steeevin88:
we can't run any external service in unit tests (Redis, Mongo, Kafka, anything else...) so they'll need to be mocked. we'll need to make a general interface to locations that use each respective client... that allows us to pass in a mock client instead

steeevin88:
let's do each in a different PR, so one for Redis, then another for Mongo, etc... 

order is up to you, I'd guess Redis is the easiest though b/c it's really only used in a few areas

PR #80, opened 2026-04-20
Added Redis store unit tests with miniredis

maernest04:
For issue #54 and subissue #75

## Problem
Redis seat/headcount logic is critical to registration correctness, but Redis behavior coverage was minimal.
Without tests around seat decrementing and full-capacity behavior, regressions could slip through.

## Solution
Added Redis-focused tests in pkg/db using miniredis to validate store behavior without relying on an external Redis service.

## Covered scenarios include:
- set/get event headcount
- successful seat take with decrement
- full event seat rejection
- missing headcount key error path
- seat release increment
- concurrent seat-take no-oversell behavior
- key format helper validation

## Files changed
- pkg/db/redis_test.go
- go.mod + go.sum
- pkg/db/redis_keys_test.go
- cmd/server/main.go
- pkg/handlers/event.go
- pkg/registration/consumer.go

## How to run tests
From project root (SCEvents):

`go test ./...`

Optional package-only run:

`go test -v ./pkg/db`

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -46,6 +46,11 @@ func main() {
 
 	ctx, cancel := context.WithCancel(context.Background())
 
+	redisStore := db.NewRedisStore(db.RedisClient())
+	stores := &db.Stores{
+		Redis: redisStore,
+	}
+
 	producer := registration.NewProducer(
 		[]string{cfg.KafkaBroker},
 		cfg.KafkaTopic,
@@ -55,7 +60,10 @@ func main() {
 		[]string{cfg.KafkaBroker},
 		cfg.KafkaTopic,
 		cfg.KafkaGroupID,
+		stores,
 	)
+
+	eventHandler := handlers.NewEventHandler(stores)
 
 	var wg sync.WaitGroup
 	wg.Add(1)
@@ -80,18 +88,18 @@ func main() {
 
 	events := r.Group("/events")
 	{
-		events.GET("/", handlers.GetEvents)
-		events.GET("/:id", handlers.GetEventByID)
-		events.GET("/registrations/:request_id", handlers.GetRegistrationStatus)
+		events.GET("/", eventHandler.GetEvents)
+		events.GET("/:id", eventHandler.GetEventByID)
+		events.GET("/registrations/:request_id", eventHandler.GetRegistrationStatus)
 
 		protected := events.Group("/")
 		protected.Use(middleware.RequireAuth(middleware.MembershipStateNonMember, cfg.ClientAPIURL))
 		{
-			protected.POST("/", handlers.CreateEvent)
-			protected.POST("/:id/register", handlers.RegisterForEvent(producer))
-			protected.POST("/:id/waitlist", handlers.JoinEventWaitlist)
-			protected.DELETE("/:id", handlers.DeleteEventByID)
-			protected.PATCH("/:id", handlers.UpdateEventByID)
+			protected.POST("/", eventHandler.CreateEvent)
+			protected.POST("/:id/register", eventHandler.RegisterForEvent(producer))
+			protected.POST("/:id/waitlist", eventHandler.JoinEventWaitlist)
+			protected.DELETE("/:id", eventHandler.DeleteEventByID)
+			protected.PATCH("/:id", eventHandler.UpdateEventByID)
 		}
 	}
 
--- a/pkg/db/redis.go
+++ b/pkg/db/redis.go
@@ -12,16 +12,14 @@ var (
 	redisClient *redis.Client
 )
 
-
-// Connect initializes the global Redis client using the provided address.
-
+// ConnectRedis initializes the global Redis client using the provided address.
 func ConnectRedis(addr string) error {
 	redisClient = redis.NewClient(&redis.Options{
-		Addr: addr,
-		Password: "", 
-		DB: 0, 
+		Addr:     addr,
+		Password: "",
+		DB:       0,
 	})
-	
+
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 	// Test the connection with a ping
@@ -33,7 +31,7 @@ func ConnectRedis(addr string) error {
 	return nil
 }
 
-// Disconnect closes the global Redis client if it has been initialized.
+// DisconnectRedis closes the global Redis client if it has been initialized.
 func DisconnectRedis() error {
 	if redisClient == nil {
 		return nil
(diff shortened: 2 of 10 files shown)


cmd/server/main.go
```diff
@@ -55,14 +49,10 @@ func main() {
 		[]string{cfg.KafkaBroker},
 		cfg.KafkaTopic,
 		cfg.KafkaGroupID,
+		redisStore,
```
steeevin88: hmm, i think this all works but the immediate gripe i think of is that if we had a bunch of services (we have 3 right now)... then we're going to pass in a LOT of arguments in many places

let's abstract a little: let's define a struct in the `db` package for the stores that we'll use:

```go

package db

type Stores struct {
	Redis RedisStore
	// Mongo MongoStore 
	// Kafka KafkaClient
}
```

and then here, in `main.go` we can add the stores to a new stores object:

```go
stores := &db.Stores{
		Redis: redisStore,
	}
```

and then pass the services as a single store object and, thus, a single argument. as opposed to one argument for multiple services

```go
consumer := registration.NewConsumer(
		[]string{cfg.KafkaBroker},
		cfg.KafkaTopic,
		cfg.KafkaGroupID,
		stores, // ---> NEW: Pass the registry
	)

eventHandler := handlers.NewEventHandler(stores)
```

steeevin88: i gemini`ed this solution + tried to reword hopefully it makes sense lol

cmd/server/main.go
```diff
@@ -78,75 +68,25 @@ func main() {
 		})
 	})
 
+	eventHandler := handlers.NewEventHandler(redisStore)
```
steeevin88:
```suggestion
	eventHandler := handlers.NewEventHandler(stores)
```

pkg/handlers/event.go
```diff
@@ -18,8 +18,16 @@ import (
 
 const dateLayout = "2006-01-02"
 
+type EventHandler struct {
+	redis db.RedisStore
```
steeevin88:
```suggestion
	stores *db.Stores
```

pkg/handlers/event.go
```diff
@@ -18,8 +18,16 @@ import (
...
+}
+
+func NewEventHandler(redisStore db.RedisStore) *EventHandler {
+	return &EventHandler{redis: redisStore}
+}
```
steeevin88:
```suggestion
func NewEventHandler(stores *db.Stores) *EventHandler {
	return &EventHandler{stores: stores}
}
```

pkg/handlers/event.go
```diff
@@ -107,7 +115,7 @@ func CreateEvent(c *gin.Context) {
 		return
 	}
 
-	if err := db.SetEventHeadcount(createdEvent.ID, createdEvent.MaxAttendees); err != nil {
+	if err := h.redis.SetEventHeadcount(createdEvent.ID, createdEvent.MaxAttendees); err != nil {
```
steeevin88:
```suggestion
	if err := h.stores.Redis.SetEventHeadcount(createdEvent.ID, createdEvent.MaxAttendees); err != nil {
```

pkg/handlers/event.go
```diff
@@ -229,7 +237,7 @@ func UpdateEventByID(c *gin.Context) {
 	if maxAttendees, ok := fields["max_attendees"]; ok {
 		if newMax, ok := maxAttendees.(float64); ok {
 			// Get current remaining seats from Redis
-			currentRemaining, err := db.GetEventHeadcount(id)
+			currentRemaining, err := h.redis.GetEventHeadcount(id)
```
steeevin88:
```suggestion
			currentRemaining, err := h.stores.Redis.GetEventHeadcount(id)
```

steeevin88: update other references as well

pkg/db/redis.go
```diff
@@ -8,48 +8,42 @@ import (
...
-	redisClient *redis.Client
-)
-
+// RedisStore defines the interface for all operations regarding Redis.
+type RedisStore interface {
+	SetEventHeadcount(eventID string, capacity int) error
+	GetEventHeadcount(eventID string) (int, error)
+	TryTakeEventSeat(eventID string) (bool, error)
+	ReleaseEventSeat(eventID string) error
+	IsUserRegisteredForEvent(eventID string, userID string) (bool, error)
```
steeevin88: ngl im not a golang expert but gemini says this:

```
In pkg/db/redis.go, the new interface methods create a fresh context.Background() internally. While this works, it's highly recommended in Go to pass the context down from the caller.

Why this matters:
- Cancellation: If a user closes their browser mid-request, Gin cancels its context. If Redis uses that context, it instantly aborts the query, saving server resources. context.Background() keeps grinding away blindly.
- Timeouts: If Redis hangs, passing a context with a timeout prevents the whole API from deadlocking.
```

and the solution would be to pass context as the first parameter of all the methods:
```go
type RedisStore interface {
    SetEventHeadcount(ctx context.Context, eventID string, capacity int) error
    GetEventHeadcount(ctx context.Context, eventID string) (int, error)
    TryTakeEventSeat(ctx context.Context, eventID string) (bool, error)
    ReleaseEventSeat(ctx context.Context, eventID string) error
    IsUserRegisteredForEvent(ctx context.Context, eventID string, userID string) (bool, error)
    Close() error // (Close usually doesn't need a context)
}
```

and pass context in calls:
```
// Example in CreateEvent:
err := h.stores.Redis.SetEventHeadcount(c.Request.Context(), createdEvent.ID, createdEvent.MaxAttendees)


// In consumer.go
func (c *Consumer) ProcessKafkaMessage(ctx context.Context, raw []byte) error {
    // ...
    ok, err := c.redis.TryTakeEventSeat(ctx, msg.EventID)
}
```

i think this makes sense 😅 if it's too complicated we could skip but ifl might as well do it right now as opposed to later

steeevin88 (changes requested):

steeevin88:
can you also
1. rebase
2. add the `go test` command to our CI? so our tests run everytime we push

thanks boss

pkg/db/redis.go
```diff
@@ -125,27 +117,42 @@ func TryTakeEventSeat(eventID string) (bool, error) {
...
+	key := EventHeadcountKey(eventID)
+	return redisClient.Set(ctx, key, capacity, 0).Err()
+}
+
+func GetEventHeadcount(ctx context.Context, eventID string) (int, error) {
+	if redisClient == nil {
+		return 0, fmt.Errorf("redis not connected")
+	}
+	key := EventHeadcountKey(eventID)
+	return redisClient.Get(ctx, key).Int()
```
steeevin88: shouldn't these no longer be needed? we should update all references in this PR so there's no need to worry about backwards compatibility

steeevin88 (changes requested):

pkg/db/redis.go
```diff
@@ -50,8 +48,15 @@ func RedisClient() *redis.Client {
 	return redisClient
 }
 
+type RealRedisStore struct {
+	client *redis.Client
+}
+
+func NewRedisStore(client *redis.Client) *RealRedisStore {
+	return &RealRedisStore{client: client}
+}
```
steeevin88:
```suggestion
type redisStore struct {
	client *redis.Client
}

func NewRedisStore(client *redis.Client) RedisStore {
	return &redisStore {client: client}
}
```

i dislike calling it "real" b/c it implies the existence of a "fake" client in our app, which technically isn't true b/c there is only one + the "fake" clients are used in testing (separate from the app)

let's make it a private struct instead, update all the other references please

steeevin88 (changes requested):

steeevin88 (approved):

PR #82, opened 2026-04-21
Added KafkaProducer interface and consumer/producer unit tests

maernest04:
### Problem
The Kafka producer has no interface for dependency injection, and the consumer's `processKafkaMessage` called package-level `db.*` functions directly, making it impossible to unit test without a live MongoDB instance.

### Solution
Added `KafkaProducer` interface for the producer, and `MongoStore` interface to enable refactoring the consumer away from package-level `db.*` calls.

### Files Changed

| File | Status | Description |
|---|---|---|
| `pkg/db/stores.go` | **Modified** | Added `KafkaProducer` and `MongoStore` interfaces, updated `Stores` struct |
| `pkg/registration/consumer.go` | **Modified** | Refactored `processKafkaMessage` to use `c.stores.Mongo` |
| `pkg/registration/producer_test.go` | **Modified** | Added validation tests, close edge cases, mock producer |
| `pkg/registration/consumer_test.go` | **Added** | 8 test cases covering all consumer business logic paths |

### How to Run Tests
```bash
go test ./pkg/registration/... -v -count=1
```

### DIFF
--- a/pkg/db/stores.go
+++ b/pkg/db/stores.go
@@ -16,6 +16,11 @@ type RedisStore interface {
 	Close() error
 }
 
+type KafkaProducer interface {
+	PublishRegistration(ctx context.Context, requestID string, eventID string, userID string) error
+	Close() error
+}
+
 type MongoStore interface {
 	GetEvents(ctx context.Context, startDate, endDate string) ([]models.Event, error)
 	GetEventByID(ctx context.Context, id string) (*models.Event, error)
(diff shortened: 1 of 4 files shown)


steeevin88 (approved):
lgtm but let's merge the MongoDB PR first, then rebase this PR; only the Kafka code should remain + then we can merge

steeevin88:
once this is merged, what files/packages still need unit tests? looks like consumer + producer are covered, anything else?

Issue #83, opened 2026-04-22
Refactor /db and /registration


steeevin88:
there's a lot of stuff in `pkg/db`, especially now that we're going to add tests for each file...

let's move things to their own packages in `/db`, ex. `/db/redis`

we need to make sure the git blame does NOT get overridden.. i believe the simplest way to do this is to have two commits:

1. Move all the files --> commit 1
2. Fix all the imports --> commit 2
3. Merge into main WITHOUT squashing

typically we want to squash + merge but I think squashing here would mess up the git blame (and thus code history)

let's also do this for `/registration`, we might as well

steeevin88:
closing b/c Go best practice is the existing file strucutre...

PR #84, opened 2026-04-22
Refactoring db and registration into better locations

maernest04:
For #83 

## Problem
pkg/db and pkg/registration had grown large and were harder to test/maintain as-is. We needed to reorganize package locations without losing useful git blame history.

## Solution
Moved packages first, then updated imports in a separate follow-up commit:

- Move-only commit (pkg/db → db, pkg/registration → registration)
- Import-fix commit (updated references to new package paths)
- This keeps history clean and preserves attribution; PR should be merged without squash.

## Files Changed

- db/** (moved from pkg/db/**)
- registration/** (moved from pkg/registration/**)
- cmd/server/main.go (import path updates)
- pkg/handlers/event.go (import path updates)


steeevin88 (changes requested):
1. can you rename the PR
2. oops i meant like, still in `pkg/db`. more so like this:

`pkg/db/redis` contains `redis.go`, `redis_keys_test.go`, `redis_test.go`
`pkg/db/event` contains `event.go`, `event_test.go`
etc... so `registration`, `stores` and `waitlist`, which will have their own test file as well

steeevin88 (commented):
hmm... i don't like the `db{service}` alias 😅 but it looks like there isn't a way to avoid this, i was thinking something like `db.{service}` but Go doesn't allow that

looked into this further (talked to Gemini) and the best practice for Go projects is to actually just keep `/db` with a bunch of files 😭  so i think i'll just close this PR + the issue, sorry for the unnecessary work

Issue #87, opened 2026-04-24
Max attendees allows for 0 attendees


maernest04:
Currently, an admin can create an event with a max attendees number of 0

PR #88, opened 2026-04-24
Fix max attendees validation and add max attendees testing

maernest04:
For issue #87 

## **Problem**
Currently, the application validates `max_attendees` by checking if it is strictly less than 0 (`< 0`). This erroneously allows events to be created with a maximum of `0` attendees, which is an invalid state since event capacities should be strictly greater than 0. Furthermore, there were no model-level unit tests verifying this validation logic.

## **Solution**
- Updated the `Validate()` method in `pkg/models/event.go` to enforce that `MaxAttendees` must be `> 0` (rejecting both `0` and negative values).
- Created `pkg/models/event_test.go` and added a table-driven unit test specifically targeting the `MaxAttendees` validation boundary conditions (testing positive, zero, and negative values) to prevent future regressions.

## **Files Changed**
- `pkg/models/event.go`: Updated `MaxAttendees` validation condition and error message.
- `pkg/models/event_test.go`: (New file) Added `TestEvent_Validate_MaxAttendees` unit test.

## **How to Run Tests**
To verify the changes, you can run the model unit tests from the root directory:
```bash
go test ./pkg/models/... -v
```

### DIFF
--- a/pkg/models/event_test.go
+++ b/pkg/models/event_test.go
@@ -108,6 +108,72 @@ func TestValidate(t *testing.T) {
 	}
 }
 
+func TestEvent_Validate_MaxAttendees(t *testing.T) {
+	// Setup a valid base event to isolate MaxAttendees validation
+	baseEvent := func() Event {
+		return Event{
+			Name:       "Test Event",
+			Date:       "2026-05-01",
+			Time:       "10:00",
+			Location:   "Room 101",
+			Status:     StatusDraft,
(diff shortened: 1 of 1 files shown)


steeevin88 (approved):
nit --> can u rename the PR? we shouldn't have present tense words in PR titles, let's just say "add" instead

so something like  ` Fix max attendees validation + Add max attendees testing`

PR #89, opened 2026-04-24
prevent event admins from registering for their own event

codebyemily:


steeevin88 (changes requested):
could we have a unit test to test this? not sure where it would live though, i'm guessing not in `/handlers` but i could be wrong

steeevin88 (approved):
thanks!

PR #90, opened 2026-04-24
Add end date checks

### LINKED ISSUE #48: Prevent the case where end_date is before date
After adding end_date to the Event struct, we should enforce validation to prevent cases where an event’s end_date occurs before its date.

Currently, the backend accepts and stores end_date without validating its relationship to date, which could lead to inconsistent or invalid event data.

Proposed Change:

Add validation logic in the backend (likely in the create and update handlers) to ensure:
end_date >= date
Return a 400 Bad Request with an appropriate error message if the validation fails

For more information, please reference #43 discussion.

codebyemily:
resolves #48

### DIFF
--- a/pkg/models/event.go
+++ b/pkg/models/event.go
@@ -3,6 +3,7 @@ package models
 import (
 	"fmt"
 	"strings"
+	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/models/event.go
```diff
@@ -130,6 +131,20 @@ func (e *Event) Validate() error {
...
+		date, err := time.Parse("2006-01-02", e.Date)
+		if err != nil {
+			return fmt.Errorf("date is not a valid date (expected YYYY-MM-DD)")
+		}
+		endDate, err := time.Parse("2006-01-02", e.EndDate)
+		if err != nil {
+			return fmt.Errorf("end_date is not a valid date (expected YYYY-MM-DD)")
+		}
+		if endDate.Before(date) {
+			return fmt.Errorf("end_date must be on or after date")
```
steeevin88: nit: can we say on or after **start** date, just to be more specific

steeevin88 (changes requested):
small stuff

1. can we add unit tests for this too 😅 
2. can you link the issue that this solves in the PR descrtion? ex "resolves #", just so the issue closes automatically + we can reference it here if we look back

also nit can you capitalize the PR title, thanks

steeevin88 (approved):
nice!

PR #91, opened 2026-04-24
Fix unlimited attendees logic

### LINKED ISSUE #71: Fix unlimited attendees logic
https://github.com/SCE-Development/SCEvents/blob/993296efbb7b32e88d395675919ef6cd3f6a7a1f/pkg/models/event.go#L119-L121

but on the frontend:

https://github.com/SCE-Development/Clark/blob/eee79c11aa1d0be5447c30a9477388ed48d5dd0d/src/Pages/Events/CreateEventPage.js#L10-L11

as a result creating an event with unlimted attendees doesnt' work, we need to fix this

1. ideally we have a better value than -1
2. make sure this works in all the places that touch this headcount value. for instance registration checks if there are still headcount (headcount > 0)... but in Redis we would have headcount as -1

this being said... if an event has unlimited registration, we probably don't even want to send the request to Kafka / write the event's headcount to Redis. this is another solution path.

ideally we go with solution 2 but feel free to solve this however you want

miiiyyi:
Closes #71 
- skip writing event headcount to Redis for unlimited events
- add Redis headcount deletion support for events that become unlimited
- update event PATCH logic to correctly handle:
  - finite -> unlimited
  - unlimited -> finite
  - finite -> finite
- update Kafka registration consumer to skip Redis seat-taking for unlimited events

Update:
- add unit tests for max-attendees headcount transition logic

### DIFF
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"context"
 
 	"github.com/SCE-Development/SCEvents/pkg/db"
 	"github.com/SCE-Development/SCEvents/pkg/models"
@@ -127,6 +128,8 @@ func (h *EventHandler) CreateEvent(c *gin.Context) {
 		return
 	}
 
+	// max_attendees == -1 means the event has unlimited attendees,
+	// so we skip Redis headcount tracking for that case.
 	if createdEvent.MaxAttendees != -1 {
 		if err := h.stores.Redis.SetEventHeadcount(c.Request.Context(), createdEvent.ID, createdEvent.MaxAttendees); err != nil {
 			c.JSON(http.StatusInternalServerError, gin.H{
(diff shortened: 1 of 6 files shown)


pkg/handlers/event.go
```diff
@@ -127,11 +127,13 @@ func (h *EventHandler) CreateEvent(c *gin.Context) {
 		return
 	}
 
-	if err := h.stores.Redis.SetEventHeadcount(c.Request.Context(), createdEvent.ID, createdEvent.MaxAttendees); err != nil {
-		c.JSON(http.StatusInternalServerError, gin.H{
-			"error": "failed to store event headcount",
-		})
-		return
+	if createdEvent.MaxAttendees != -1 {
```
steeevin88, line 133: can we add a small comment above this if block saying if max attendees is -1 it means unlimited attendees? just so someone else looking here wouldn't get confused

steeevin88 (changes requested):
lgtm, could we add unit tests for this logic? ex. updating an event's headcount from finite to infinite, infinite to finite, etc...

i think this should be possible, if not lmk

pkg/handlers/event.go
miiiyyi: I moved the existing max-attendees Redis sync logic into a helper method for testability, then added unit tests for finite -> unlimited, unlimited -> finite, and finite -> finite transitions.

steeevin88 (approved):
nice! i'm gonna refactor this slightly in #99 , I didn't realize the existing mocks were exportable 😅

PR #92, opened 2026-04-24
Fix Update Endpoint

KhoaNguyen706:
Admin of event created the event:
image
Try edit registration and new type, such as public status, visibility, and minium visible role, attendance:
image
Update successfully!:
image
-Use another account to test in order to see it changed or not:
image
It work well once changing the events, and use another account to see it:
image

### DIFF
--- a/pkg/models/event.go
+++ b/pkg/models/event.go
@@ -348,7 +348,49 @@ func (e *Event) ApplyPatch(fields map[string]interface{}) error {
 			}
 			e.WaitlistSize = int(n)
 			
-		case "admins", "registration_form":
+		case "registration_form":
+			arr, ok := value.([]interface{})
+			if !ok {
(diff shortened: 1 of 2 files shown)


pkg/models/event.go
```diff
@@ -333,7 +333,49 @@ func (e *Event) ApplyPatch(fields map[string]interface{}) error {
...
+						if s, ok := opt.(string); ok {
+							q.AnswerOptions = append(q.AnswerOptions, s)
+						}
+					}
+				}
+				questions = append(questions, q)
+			}
+			e.RegistrationForm = questions
+
+		case "admins":
```
steeevin88, line 393: does this mean admins aren't editable? we should probably add this in a new PR / make an issue for this

steeevin88 (approved):
nice

PR #94, opened 2026-04-25
Block registration for closed events

trista-chen-29:
### Problem
- Users could register for events even when `status = closed`
- Kafka consumer could still process and potentially accept those registrations

### Changes
- Added validation in `RegisterForEvent` handler:
  - Reject requests when `event.status == closed`
- Added validation in Kafka consumer:
  - Reject pending registrations if event becomes closed before processing
- Introduced new rejection reason:
  - `event_closed`
- Added unit test:
  - `TestProcessKafkaMessage_EventClosed`

#### Manual Testing
- Open event → registration accepted
- Full event → rejected (`capacity_full`)
- Closed event → blocked at API level

Example response:
```json
{"error":"registration is closed for this event"}
```

- Kafka consumer correctly rejects with:
   - `decision_reason = event_closed`

#### Automated Tests
```bash
go test ./...
```
All tests passing.

### DIFF
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -330,6 +330,13 @@ func (h *EventHandler) RegisterForEvent(producer *registration.Producer) gin.Han
 			return
 		}
 
+		if ev.Status == models.StatusClosed {
+			c.JSON(http.StatusForbidden, gin.H{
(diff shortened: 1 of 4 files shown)


steeevin88 (approved):
awesome, thanks for testing too

PR #101, opened 2026-04-25
Add exportable mocks

### LINKED ISSUE #99: Support exportable mock services (for unit tests)
we'll have various files that will interact with redis, mongo, kafka, etc... and in the unit tests, they should interat with MOCKS of these services

while #54 adds this, we need to ensure that the mocked services are IMPORTABLE... which they currently aren't. refactor the mock services to be importable + update any existing references

steeevin88:
**CHANGES**
- adds a new `mocks` directory containing mocks for each of the three core services in SCEvents (Redis, Kafka, Mongo)
- updated tests across project to utilize mock services if needed (ex. tests from #71)

**TESTING**
- CI passes which runs `go test`

resolves #99


c1ndytat (approved):
what's coffee backwards

Issue #95, opened 2026-04-25
Support retrieving event attendees


steeevin88:
when we open an event card, we should get more information on the attendees of an event, ex.:
- number of attendees going
- if the attendee list if public, the names of the attendees going

however, our setup currently doesn't support this:
- Redis is the source of truth for event headcount
- I'm pretty sure we don't actually write this headcount to the event EVER... which is fine --> our source of truth is the `registrations` collection

let's support this, i drafted a gameplan with copilot but feel free to modify if needed:

1. Querying: Create a new query function in `db/registration.go` that finds all documents in the registrations collection where `event_id` equals the requested ID and status is "accepted".
2. Endpoint: Create a new GetEventAttendees endpoint in `handlers/event.go` that calls this DB function.

future point-->  Privacy Logic: Regarding public vs. private attendee lists:You can add a new boolean field to the Event struct (e.g., IsAttendeeListPublic bool).
If the event's attendee list is set to public, your endpoint can map over the accepted Registrant documents and return their names to the client. Keep in mind not to leak private info like email addresses.
Alternatively, if the privacy is at the user/registrant level (e.g. "I want my own name hidden"), you can add a HideNameFromPublicList boolean to the RegistrationPayload and RegistrationRequest. Then, when an event's attendee list is requested, the endpoint returns names for those who opted-in, and generic "Anonymous Attendee" ..

on second thought for this point, let's just not do this. we'd have to indicate for people trying to register for events if their registration will be public... which is a whole other ordeal. so let's just skip it for now, i'll open a new issue for this: https://github.com/SCE-Development/SCEvents/issues/96

PR #104, opened 2026-04-26
Add admin event attendance count

maernest04:
# For PR #95

## Problem

SCEvents did not expose accepted attendee counts for event cards. Unlimited events were also treated as full because Redis stored no-limit capacity as `-1`, but seat reservation treated values `<= 0` as unavailable.

## Solution

- Added accepted-registration counting from Mongo.
- Added `GET /events/:id/attendance`.
- Restricted attendance count access to event admins/site admins.
- Initialized Mongo in `Stores`.
- Completed registrations synchronously when capacity allows.
- Updated Redis seat logic so `-1` means unlimited.
- Added tests for attendee counting and unlimited capacity behavior.

## Files Changed

- `SCEvents/cmd/server/main.go`
- `SCEvents/pkg/db/registration.go`
- `SCEvents/pkg/db/stores.go`
- `SCEvents/pkg/db/redis.go`
- `SCEvents/pkg/db/redis_test.go`
- `SCEvents/pkg/db/registration_test.go`
- `SCEvents/pkg/handlers/event.go`
- `SCEvents/pkg/handlers/event_test.go`
- `SCEvents/pkg/registration/consumer_test.go`

### Unlimited Events Were Treated As Full

`max_attendees = -1` was stored in Redis, but seat reservation treated `-1` as full.

Solution: treat negative headcount as unlimited and do not decrement.

### Store Wiring Was Incomplete

The new attendance handler used `stores.Mongo`, but server startup only initialized Redis in `Stores`.

Solution: initialize Mongo in `Stores`.

### Pending Registration Did Not Affect Count

The attendee count only counts accepted registrations.

Solution: complete registration synchronously when capacity allows.

## Possible Follow-Up Issues

- Add public attendee names/lists with consent UX.
- Refactor DB access to consistently use injected stores instead of mixing global DB functions and `stores.Mongo`.

### DIFF
--- a/pkg/db/redis.go
+++ b/pkg/db/redis.go
@@ -96,12 +96,30 @@ if not current then
 end
 
 current = tonumber(current)
-if current <= 0 then
+if current < 0 then
+	return 1
+end
+
+if current == 0 then
 	return 0
 end
 
 redis.call("DECR", KEYS[1])
 return 1
+`)
+
+var releaseSeatScript = redis.NewScript(`
+local current = redis.call("GET", KEYS[1])
+if not current then
+	return -2
+end
+
+current = tonumber(current)
+if current < 0 then
+	return current
+end
+
+return redis.call("INCR", KEYS[1])
 `)
 
 func (s *redisStore) TryTakeEventSeat(ctx context.Context, eventID string) (bool, error) {
@@ -130,7 +148,7 @@ func (s *redisStore) ReleaseEventSeat(ctx context.Context, eventID string) error
 		return fmt.Errorf("redis not connected")
 	}
 	key := EventHeadcountKey(eventID)
-	return s.client.Incr(ctx, key).Err()
+	return releaseSeatScript.Run(ctx, s.client, []string{key}).Err()
 }
 
 func (s *redisStore) IsUserRegisteredForEvent(ctx context.Context, eventID string, userID string) (bool, error) {
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -98,6 +98,43 @@ func (h *EventHandler) GetEventByID(c *gin.Context) {
 	c.JSON(http.StatusOK, event)
 }
 
+func (h *EventHandler) GetEventAttendanceSummary(c *gin.Context) {
+	eventID := strings.TrimSpace(c.Param("id"))
+	if eventID == "" {
+		c.JSON(http.StatusBadRequest, gin.H{
+			"error": "event id is required",
+		})
+		return
+	}
+
+	_, err := h.stores.Mongo.GetEventByID(c.Request.Context(), eventID)
+	if err != nil {
+		if errors.Is(err, mongo.ErrNoDocuments) {
+			c.JSON(http.StatusNotFound, gin.H{
+				"error": "event not found",
+			})
+			return
+		}
+		c.JSON(http.StatusInternalServerError, gin.H{
+			"error": "failed to fetch event",
+		})
+		return
+	}
+
+	attendeeCount, err := h.stores.Mongo.CountAcceptedRegistrationsForEvent(c.Request.Context(), eventID)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{
+			"error": "failed to fetch attendance summary",
+		})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"event_id":        eventID,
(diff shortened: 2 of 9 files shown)


pkg/handlers/event_test.go
```diff
@@ -2,85 +2,169 @@ package handlers
...
+		router.ServeHTTP(w, req)
+
+		if w.Code != http.StatusForbidden {
+			t.Fatalf("expected status 403, got %d", w.Code)
+		}
+		if mongoStore.countCalled {
+			t.Fatal("expected attendee count query not to be called")
+		}
+	})
 }
```
steeevin88, line 115: i added an exportable mock client in #101 , could we use those here instead?

pkg/handlers/event.go
```diff
@@ -404,15 +415,33 @@ func (h *EventHandler) RegisterForEvent(producer *registration.Producer) gin.Han
...
+			c.JSON(http.StatusInternalServerError, gin.H{
+				"error": "failed to complete registration",
 			})
 			return
 		}
 
-		c.JSON(http.StatusAccepted, gin.H{
-			"message":    "registration request sent",
+		c.JSON(http.StatusCreated, gin.H{
+			"message":    "registration complete",
```
steeevin88: this should not have been changed i think? the registration is sent but not necessary complete b/c it has to go through the producer/consumer flow

pkg/handlers/event.go
```diff
@@ -404,15 +415,33 @@ func (h *EventHandler) RegisterForEvent(producer *registration.Producer) gin.Han
...
+				"error": "event is full",
+			})
+			return
+		}
+
+		if err := db.MarkRegistrationAccepted(created.RequestID); err != nil {
+			_ = h.stores.Redis.ReleaseEventSeat(c.Request.Context(), eventID)
+			_ = db.MarkRegistrationRejected(created.RequestID, models.ReasonInternalError)
+			c.JSON(http.StatusInternalServerError, gin.H{
+				"error": "failed to complete registration",
```
steeevin88: let's say "failed to send registration request" instead

pkg/handlers/event.go
```diff
@@ -292,11 +292,29 @@ func (h *EventHandler) UpdateEventByID(c *gin.Context) {
...
+				currentRemaining, err := h.stores.Redis.GetEventHeadcount(c.Request.Context(), id)
+				if err != nil {
+					currentRemaining = existingEvent.MaxAttendees
+				}
+
+				seatsTaken := existingEvent.MaxAttendees - currentRemaining
+				newRemaining = int(newMax) - seatsTaken
+				if newRemaining < 0 {
+					newRemaining = 0
+				}
```
steeevin88: i don't get what's going on in this snippet + there aren't any comments here (as opposed to the earlier version), can you add some comments?

could you also explain the negative logic? if `newRemaining` is < 0, why did we set it to 0?

pkg/db/redis.go
```diff
@@ -81,29 +81,39 @@ func (s *redisStore) GetEventHeadcount(ctx context.Context, eventID string) (int
...
+end
+
+current = tonumber(current)
+if current < 0 then
+	return current
+end
+
+return redis.call("INCR", KEYS[1])
+`)
+
```
steeevin88, line 124: can you double check if an event has unlimited events, that we write the headcount to redis? i thought we didn't but i could be wrong

please reply to this comment with an explanation of what happens, thanks

steeevin88 (changes requested):

pkg/db/redis.go
maernest04: When the max attendees value is set to -1, its not tracked in Redis, the incrementing gets skipped. the part where it says current < 0 is more of a safe fall back when the value we put in max attendees is negative, which now that i think about it, technically won't even happen cause we don't let the event be made if its less than -1

pkg/handlers/event.go
maernest04: This part of the code is supposed to account for when we update an event but we reduce its max attendees to less than its current attendee count. So for example if the event's original attendee limit was 100, we got 60 sign ups, then we reduce our attendee limit to 50, then we need to stop registration since we've technically hit capacity now. We set the newRemaining to 0 so redis doesn't get  confused thinking we have negative seats remaining, rather a flat 0.

steeevin88 (approved):
dope, thanks for the changes!

PR #107, opened 2026-04-27
Persist data in prod

steeevin88:
our data was being handled by Docker which is fine but i think we should prefer to mount data to volumes on the server? 

- we can switch to some other solution later if needed
- i don't think this will overload our server, doubt events will be a lot of data (in the short term at least)


c1ndytat (approved):

maernest04 (approved):
How does cereal pay its bills? With Chex.

PR #108, opened 2026-04-27
Add registration_status to events response and related tests

### LINKED ISSUE #2097: Render event cards differently based on users' registration status
self-explanatory, if someone is registered for an event they should know... or if they got rejected, it should probably say that as well

- not sure what the best way to do this is; you'll need to figure out how to get the attendees of an event, and then check if the current user is an attendee of the event --> this is probably inefficient; maybe the better approach is to make a new backend endpoint, something like "is user registered for this event". but then THAT's super inefficient because you can imagine this endpoint would be called for EVERY single event...
	- you'll need to research efficient ways to determine if a user is a registrant of an event...
	- again maybe it's a new endpoint. maybe it's modifying existing endpoints (ex. getEvents)

- the rejected logic should probably sync with #2091; if they're rejected we should let them join the waitlist if available. up to you if you want to add the rejection as part of this issue or the waitlist one

feel free to make these design decisions yourself

trista-chen-29:
Partially resolved [Render event cards differently based on users' registration status #2097](https://github.com/SCE-Development/Clark/issues/2097)

### Summary
Adds `registration_status` to event responses so the frontend can render user-specific event states.

### Changes
- Augmented `GET /events` and `GET /events/:id` to include `registration_status`
- Implemented batch lookup for user registration statuses across events
- Added waitlist integration (waitlist overrides rejected)
- Introduced `OptionalAuth` for public endpoints with user context
- Added comprehensive unit tests for:
  - registration status aggregation
  - handler responses
  - middleware behavior

## Validation
- `go test ./...` passes
- Verified via curl with valid JWT:
  - accepted → `registered`
  - no record → `none`
 
```bash
curl -H "Authorization: Bearer <WORKING_TOKEN>" \
http://localhost:8002/events/
```

For examples:

```json
[{"id":"","name":"Spring Social","date":"2026-04-20","time":"18:00","location":"Student Union","description":"Casual meetup","admins":["user123"],"registration_form":[],"max_attendees":50,"created_at":"","status":"draft","visibility":"public","waitlist_enabled":false,"registration_status":"none"}
```

```json
{"id":"test-event-1","name":"Mongo Store Wiring Test","date":"2026-04-30","time":"6:00 PM","location":"ENGR 289","description":"test event","admins":[],"registration_form":[{"id":"major","type":"textbox","question":"What is your major?","required":true}],"max_attendees":1,"created_at":"2026-04-25T00:00:00Z","status":"closed","visibility":"public","waitlist_enabled":false,"registration_status":"registered"}
```

- Confirmed correctness against MongoDB registration data

## Notes
- [Design Doc](https://docs.google.com/document/d/1wE4X-_0kccqDb6SvqMAUr_P2v80PHeJlSTlgf_-5Nsk/edit?usp=sharing)

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -53,6 +53,7 @@ func main() {
 	mongoStore := db.NewMongoStore(
 		db.GetEventsCollection(),
 		db.GetRegistrationsCollection(),
+		db.GetWaitlistCollection(),
 	)
 
 	stores := &db.Stores{
@@ -96,10 +97,12 @@ func main() {
 	})
 
 	events := r.Group("/events")
+	// public event reads can use optional auth so authenticated callers receive personalized event metadata
+	events.Use(middleware.OptionalAuth(cfg.ClientAPIURL))
 	{
+		events.GET("/registrations/:request_id", eventHandler.GetRegistrationStatus)
 		events.GET("/", eventHandler.GetEvents)
 		events.GET("/:id", eventHandler.GetEventByID)
-		events.GET("/registrations/:request_id", eventHandler.GetRegistrationStatus)
 
 		protected := events.Group("/")
 		protected.Use(middleware.RequireAuth(middleware.MembershipStateNonMember, cfg.ClientAPIURL))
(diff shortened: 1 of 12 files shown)


steeevin88 (approved):
looks fine lol thanks for making the design doc, super helpful

Issue #98, opened 2026-04-25
Allow editing admins


steeevin88:
as per changes/comments in https://github.com/SCE-Development/SCEvents/pull/92, i think admins aren't editable at the current moment

we should allow this with some restrictions:
- can admins remove themselves? remove other admins? etc...
- we might want to establish new roles

PR #109, opened 2026-04-29
Added multi admin event management

maernest04:
# PR for #98 

## **Problem**
Event creation allowed admins to create and edit their own events but there was no way for other admins to be able to edit events they didn't create. There should be a way for the event's original creator to add and remove admins who can edit their event. 

## **Solution**
- Moved `POST /events` (CreateEvent) into a new admin-protected route group requiring `MembershipStateAdmin`
- CreateEvent handler now extracts the creator's user ID from the auth context and ensures it's always included in the admins list with deduplication
- Enabled the `admins` field in `ApplyPatch` so event updates can modify the admin list with validation (type checking, dedup, non-empty)
- Added `accessLevel` to the Gin context in both `RequireAuth` and `OptionalAuth` middleware

## **Files Changed**
| File | Change |
|------|--------|
| `cmd/server/main.go` | New admin-protected route group for event creation |
| `pkg/handlers/event.go` | Creator forced into admins, admins passthrough on update |
| `pkg/handlers/event_test.go` | Tests for creator admin inclusion and dedup |
| `pkg/middleware/auth.go` | Expose `accessLevel` in context |
| `pkg/middleware/auth_test.go` | Tests for accessLevel and admin auth rejection |
| `pkg/models/event.go` | Enable admins patching with validation |
| `pkg/models/event_test.go` | Tests for admins patch cases |
| `pkg/mocks/mongo.go` | Track created event in mock for test assertions |

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -107,7 +107,6 @@ func main() {
 		protected := events.Group("/")
 		protected.Use(middleware.RequireAuth(middleware.MembershipStateNonMember, cfg.ClientAPIURL))
 		{
-			protected.POST("/", eventHandler.CreateEvent)
 			protected.GET("/:id/attendance", eventHandler.GetEventAttendanceSummary)
 			protected.GET("/:id/registrations", eventHandler.ListEventRegistrations)
 			protected.GET("/:id/registrations/:request_id", eventHandler.GetEventRegistrationByRequestID)
@@ -115,6 +114,12 @@ func main() {
 			protected.POST("/:id/waitlist", eventHandler.JoinEventWaitlist)
 			protected.DELETE("/:id", eventHandler.DeleteEventByID)
 			protected.PATCH("/:id", eventHandler.UpdateEventByID)
+		}
+
+		adminProtected := events.Group("/")
+		adminProtected.Use(middleware.RequireAuth(middleware.MembershipStateOfficer, cfg.ClientAPIURL))
+		{
+			adminProtected.POST("/", eventHandler.CreateEvent)
 		}
 	}
 
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -112,14 +112,14 @@ func main() {
 			protected.GET("/:id/registrations/:request_id", eventHandler.GetEventRegistrationByRequestID)
 			protected.POST("/:id/register", eventHandler.RegisterForEvent(producer))
 			protected.POST("/:id/waitlist", eventHandler.JoinEventWaitlist)
-			protected.DELETE("/:id", eventHandler.DeleteEventByID)
-			protected.PATCH("/:id", eventHandler.UpdateEventByID)
 		}
 
-		adminProtected := events.Group("/")
-		adminProtected.Use(middleware.RequireAuth(middleware.MembershipStateOfficer, cfg.ClientAPIURL))
+		officerProtected := events.Group("/")
+		officerProtected.Use(middleware.RequireAuth(middleware.MembershipStateOfficer, cfg.ClientAPIURL))
 		{
-			adminProtected.POST("/", eventHandler.CreateEvent)
+			officerProtected.POST("/", eventHandler.CreateEvent)
+			officerProtected.DELETE("/:id", eventHandler.DeleteEventByID)
+			officerProtected.PATCH("/:id", eventHandler.UpdateEventByID)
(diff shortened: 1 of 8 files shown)


pkg/models/event_test.go
```diff
@@ -383,13 +383,87 @@ func TestApplyPatch(t *testing.T) {
...
+			}
+			if err.Error() != tt.wantErr {
+				t.Fatalf("expected error %q, got %q", tt.wantErr, err.Error())
+			}
+		})
+	}
+}
+
+func sameStringSlice(a, b []string) bool {
+	if len(a) != len(b) {
```
steeevin88: this is helper just used once? can we just do it inline then

or maybe `reflect.DeepEqual(a, b)`? this is what gemini says

cmd/server/main.go
```diff
@@ -107,13 +107,18 @@ func main() {
...
 			protected.POST("/:id/register", eventHandler.RegisterForEvent(producer))
 			protected.POST("/:id/waitlist", eventHandler.JoinEventWaitlist)
 			protected.DELETE("/:id", eventHandler.DeleteEventByID)
 			protected.PATCH("/:id", eventHandler.UpdateEventByID)
 		}
+
+		adminProtected := events.Group("/")
+		adminProtected.Use(middleware.RequireAuth(middleware.MembershipStateOfficer, cfg.ClientAPIURL))
+		{
+			adminProtected.POST("/", eventHandler.CreateEvent)
```
steeevin88: good call

i suppose update shouldn't be here since, in theory, a member could be an event's admin + would need access to update the event

pkg/handlers/event_test.go
```diff
@@ -114,6 +115,138 @@ func TestGetEventAttendanceSummary(t *testing.T) {
...
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
```
steeevin88: same here, let's remove this and use a built-in option?

steeevin88 (changes requested):
small nits

cmd/server/main.go
steeevin88: hm kind of nitpicky, but is "admin protected" the best label here? maybe `officerProtected` instead? since the level is officers + above? 

if this meant "event admin" protected though, i would still argue for  `officerProtected` instead b/c, in theory, event admins could be officers, admins, etc...

pkg/handlers/event.go
```diff
@@ -263,6 +273,29 @@ func (h *EventHandler) CreateEvent(c *gin.Context) {
 	c.JSON(http.StatusCreated, createdEvent)
 }
 
+func ensureAdminIncluded(admins []string, userID string) []string {
```
steeevin88: another nitpick, maybe let's be more specific with this function name? this function ensures that the event creator is included in the event's list of admins, so maybe something like `ensureCreatorIsEventAdmin`, or maybe `ensureEventAdminsContainsCreator`

steeevin88: on second look this function does more than just check if the creator is an event admin, and this sort of gets lost if our function name is `ensureCreatorIsEventAdmin`

can we split this into two functions? ideally `sanitizeAdmins` and `ensureCreatorIsEventAdmin`

so then above would be 
```
event.Admins = sanitizeAdmins(event.Admins)
event.Admins = ensureCreatorIsEventAdmin(event.Admins, creatorID)
```

pkg/handlers/event.go
```diff
@@ -400,6 +433,10 @@ func (h *EventHandler) UpdateEventByID(c *gin.Context) {
 		fields["minimum_visible_role"] = updatedEvent.MinimumVisibleRole
 	}
 
+	if _, ok := fields["admins"]; ok {
+		fields["admins"] = updatedEvent.Admins
```
steeevin88: sanitizing is probably most important here? as opposed to in `createEvent`

let's export the sanization logic, this line should be something like:

```
if _, ok := fields["admins"]; ok {
    fields["admins"] = sanitizeAdmins(updatedEvent.Admins)
}
```

steeevin88 (changes requested):
almost there 😄

steeevin88 (approved):
nice!

PR #110, opened 2026-04-29
Create integration and load tests

### LINKED ISSUE #56: Create load tests
this project was meant to be able to handle thousands of concurrent registration requests. while we realistically will never hit this in production, we should still have tests to ensure that system can support this

we can create what's referred to as **load tests**, which is as the name suggests; tests that check if a service can handle large loads of requests (in this case, registration)

we should probably use a testing framework like `k6` to do this. let's try to flood our service with thousands of registration requests to a handful of events... and see if our service can handle it properly

- not sure how the environment would work... figure out if we need the docker compose running
- we should probably have all related files in a new top-level directory... maybe `tests`? and then `tests/load_testing`

ultimately load tests will bring really valuable insights as to if our project would actually performing as intended at scale

### LINKED ISSUE #57: Create integration/E2E tests
manually testing everything can be super tedious; we could have a single command that:
1. spins up all necessary containers
2. has mock event data to use
3. registers users to these events
4. create events
5. etc....

basically rather than manually testing everything, a command does it for us. it would be super cool if it was something like "test registration flow", or maybe "test event creation flow"...

a few notes:
- biggest issue is Clark... we'll have to research how to get around Clark / if we should bring it into the integration test requirements
- at Tesla we used `make` commands, ex. `make integration` would run the integration test. maybe we could do something similar, ex. `make integration_registration` runs the registration flow, `make integration_creationg` runs the creation flow... those commands are really long but you get the idea
- #56 adds load testing. ideally these both live under a top-level directory, ex. `/testing`... so `testing/integration` and `testing/load`

codebyemily:
Resolves #57 
Resolves #56 

This PR introduces a testing framework for SCEvents, covering both:

- Integration testing for correctness of core event and registration flows via Make targets.
- High-concurrency load testing with k6 to validate system behavior under stress.

It also adds a repeatable, Docker-based local testing setup with standardized commands for running and tearing down tests.

## Directory layout
testing/
  ├── Makefile                                  # make integration_reg / integration_event / integration_all / load_reg
  ├── integration/
  │   ├── go.mod                                # separate module, stdlib only
  │   ├── mock_clark/
  │   │   ├── main.go                           # fake Clark: echoes Bearer token as user _id
  │   │   └── Dockerfile
  │   ├── docker-compose.integration.yml        # adds mock-clark, overrides CLIENT_API_URL
  │   ├── main_test.go                          # TestMain: waits for /ping
  │   ├── helpers_test.go                       # helper funcs for integration tests
  │   ├── registration_test.go                  # registration flow coverage
  │   └── creation_test.go                      # create/retrieve/update/delete coverage

tests/
  └── load_testing/
      ├── docker-compose.load.yml               # load-test compose override
      ├── register_flood.js                     # k6 registration flood scenario
      └── README.md                             # load-test usage and thresholds

## Auth strategy in tests

`mock_clark` accepts any `Bearer <token>` and returns:

```json
{"_id":"<token>","accessLevel":1}
```

Each test uses a unique token (for example, `user-basic-reg`), so user identity is controlled by the token value without requiring a real Clark instance.

### Integration testing
make integration_reg
make integration_event
make integration_all

### Load testing 

make load_reg

## Teardown

Integration stack:

docker compose -f ../docker-compose.dev.yml -f integration/docker-compose.integration.yml down

Load-test stack:

docker compose -f ../docker-compose.dev.yml -f ../tests/load_testing/docker-compose.load.yml down

Containers stay running after each command so you can inspect state before teardown.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,33 @@
+name: Integration Tests
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+jobs:
+  integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      # Compose can be slow on runners (Kafka + topic init + server build)
+      INTEGRATION_READY_TIMEOUT: 120s
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      # make integration = docker compose up + go test -tags=integration (see testing/Makefile)
+      - name: Run integration tests
+        run: make integration
+
+      - name: Tear down Compose
+        if: always()
+        run: make teardown-integration
--- a/testing/Makefile
+++ b/testing/Makefile
@@ -1,0 +1,60 @@
+# Paths follow this Makefile location (not cwd), so you can run:
+#   make -f testing/Makefile integration
+# from the repo root, or from anywhere via absolute -f path.
+_TESTING_ABS := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+REPO_ROOT     := $(abspath $(_TESTING_ABS)/..)
+COMPOSE_BASE     := $(REPO_ROOT)/docker-compose.dev.yml
+COMPOSE_INT      := $(REPO_ROOT)/testing/integration/docker-compose.integration.yml
+COMPOSE_LOAD     := $(REPO_ROOT)/testing/load/docker-compose.load.yml
+INTEGRATION_DIR  := $(REPO_ROOT)/testing/integration
+LOAD_DIR         := $(REPO_ROOT)/testing/load
+
+# Default to running all integration tests if no filter is provided
+TEST ?= all
+
+.PHONY: integration load teardown teardown-integration teardown-load _up_integration _up_load
+
+# Tear down Compose stacks (absolute -f paths; works from any cwd).
+teardown-integration:
+	docker compose --project-directory "$(REPO_ROOT)" -f $(COMPOSE_BASE) -f $(COMPOSE_INT) down
+
+teardown-load:
+	docker compose --project-directory "$(REPO_ROOT)" -f $(COMPOSE_BASE) -f $(COMPOSE_LOAD) down
+
+teardown: teardown-integration teardown-load
+
+# Run integration tests.
+# Usage:
+#   make integration                  (runs full suite)
+#   make integration TEST=register    (registration flow only)
+#   make integration TEST=create      (event creation flow only)
+integration: _up_integration
+ifeq ($(TEST),register)
+	cd $(INTEGRATION_DIR) && go test -tags=integration -v -run TestRegistration ./...
+else ifeq ($(TEST),create)
+	cd $(INTEGRATION_DIR) && go test -tags=integration -v -run TestEventCreation ./...
+else
+	cd $(INTEGRATION_DIR) && go test -tags=integration -v ./...
+endif
+
+# Flood the registration endpoint with concurrent requests via k6
+load: _up_load
+	k6 run $(LOAD_DIR)/register.js
+
+# ── internal targets ──────────────────────────────────────────────────────────
+
+# Spin up the stack with mock Clark for integration tests (leaves containers running)
+_up_integration:
+	docker compose \
+		--project-directory "$(REPO_ROOT)" \
+		-f $(COMPOSE_BASE) \
+		-f $(COMPOSE_INT) \
+		up -d --build
+
+# Spin up the stack with mock Clark for load tests (leaves containers running)
+_up_load:
+	docker compose \
+		--project-directory "$(REPO_ROOT)" \
+		-f $(COMPOSE_BASE) \
+		-f $(COMPOSE_LOAD) \
+		up -d --build
--- a/testing/testing.md
+++ b/testing/testing.md
@@ -1,0 +1,26 @@
+# Testing Guide
+
+## Usage
+
+To run integration tests:
+
+```bash
+make integration
+make integration TEST=create   # runs event creation flow
+make integration TEST=register # runs registration flow
+```
+
+Must install K6 to run load tests.
+To run load tests:
+
+```bash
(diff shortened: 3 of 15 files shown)


testing/Makefile
```diff
@@ -0,0 +1,40 @@
...
+integration_event: _up_integration
+	cd $(INTEGRATION_DIR) && go test -v -run TestEventCreation ./...
+
+# Run both integration flows
+integration_all: _up_integration
+	cd $(INTEGRATION_DIR) && go test -v ./...
+
+# Flood the registration endpoint with thousands of concurrent requests via k6
+load_reg: _up_load
+	k6 run $(LOAD_DIR)/register_flood.js
```
steeevin88: a bit nitpicky but could we do something like this instead:
```
# Default to running all tests if no argument is provided
TEST ?= all

.PHONY: integration load_reg _up_integration _up_load

# Run integration tests. 
# Usage:
#   make integration             (runs all)
#   make integration TEST=register (runs only registration flow)
#   make integration TEST=create (runs only event creation flow)
integration: _up_integration
ifeq ($(TEST),register)
	cd $(INTEGRATION_DIR) && go test -v -run TestRegistration ./...
else ifeq ($(TEST),create)
	cd $(INTEGRATION_DIR) && go test -v -run TestEventCreation ./...
else
	cd $(INTEGRATION_DIR) && go test -v ./...
endif
```

note that `reg` changed to `register` just to be more explicit + same with `event` to `create`

reason being i'd like for `make integration` to be the default command to run our integration suite (as opposed to an `_all`), and by default we'd want to run all of our suites anyway

but in the (probably rare) case we do want only a certain flow, we can specify what we want in what's basically a separate `TEST={test_type}` value

(also note - i think it'd be BEST If we did something like `make integration {optional_test_type}`, ex. `make integration create`, but from what Gemini said it's a bit hard to do that lol)

testing/testing.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Auth strategy in tests
+
+`mock_clark` accepts any `Bearer <token>` and returns:
+
+```json
+{"_id":"<token>","accessLevel":1}
+```
+
+Each test uses a unique token (for example, `user-basic-reg`), so user identity is controlled by the token value without requiring a real Clark instance.
```
steeevin88: so from my understanding all the mock Clark service does is return the auth response? just want to confirm

testing/Makefile
steeevin88: oop and can we make the load test --> `make load`

testing/testing.md
```diff
@@ -0,0 +1,59 @@
...
+## Commands (run from `testing/`)
+
+### Integration testing
+make integration_reg
+make integration_event
+make integration_all
+
+### Load testing 
+
+make load_reg
```
steeevin88: let's update these too after earlier comment is resolved

testing/testing.md
```diff
@@ -0,0 +1,59 @@
...
+
+`mock_clark` accepts any `Bearer <token>` and returns:
+
+```json
+{"_id":"<token>","accessLevel":1}
+```
+
+Each test uses a unique token (for example, `user-basic-reg`), so user identity is controlled by the token value without requiring a real Clark instance.
+
+## Commands (run from `testing/`)
```
steeevin88: would it be possible to run this from anywhere? just a nit, if it's hard then no worries

testing/testing.md
```diff
@@ -0,0 +1,59 @@
...
+
+Integration stack:
+
+docker compose -f ../docker-compose.dev.yml -f integration/docker-compose.integration.yml down
+
+Load-test stack:
+
+docker compose -f ../docker-compose.dev.yml -f ../tests/load_testing/docker-compose.load.yml down
+
+Containers stay running after each command so you can inspect state before teardown.
```
steeevin88: could we condense in file a little bit? let's have the first section be a `Usage` section where we have subsections in the following order
- commands to run
- brief sentence noting the containers will be spun up
- teardown

ex.
```
## Usage
To run integration tests:

codeblock
make integration
make integration TEST=create  # runs event creation flow
make integration TEST=register # runs event registration flow
codeblock

These commands will spin up containers.... so you can inspect state...

To teardown:
codeblock
the stuff you have here
codeblock
```

steeevin88 (changes requested):
small nitpicky stuff, i think the general flow looks good though!

testing/testing.md
codebyemily: Yes, the mock Clark's …/api/Auth/verify target forwards the Authorization header and expects 200 with JSON containing _id (who the user is) and accessLevel (membership tier).

testing/testing.md
codebyemily: i updated it so it can run from anywhere within the root repo

steeevin88 (changes requested):
a few things:

1. can you add `make integration` to a new github pipeline? let's add it to our CI (no need for the load tests, just integration)
2. when i run `make load` locally, i get this:
```
 ✔ Container kafka                  Healthy                                                                             0.5s 
 ✔ Container scevents-redis-1       Running                                                                             0.0s 
 ✔ Container scevents-mongodb-1     Healthy                                                                             0.5s 
 ✔ Container scevents-mock-clark-1  Healthy                                                                             0.5s 
 ✔ Container kafka-init             Started                                                                             0.6s 
 ✔ Container scevents-server-1      Running                                                                             0.0s 
k6 run /Users/steeevin/projects/SCE/SCEvents/testing/load/register.js
make[1]: k6: No such file or directory
make[1]: *** [load] Error 1
make: *** [load] Error 2
```

can we streamline this so we don't have to install k6 locally? but if the only solution is to make a docker container that runs `k6`... then let's just add an instruction in the `testing.md` that says to install `k6` to run load tests lol

3. should we remove `integration.test.exe`? i think we don't need this but i could be mistaken

steeevin88 (approved):
very cool stuff 😄 integration tests will be super helpful in ensuring changes to SCEvents don't break the core application logic, and having a CI pipeline that runs the integration tests on every change improves DX a lot

thanks!

.github/workflows/integration.yml
```diff
@@ -0,0 +1,33 @@
+name: Integration tests
```
steeevin88: can you capitalize tests 😅 being nitpicky

PR #111, opened 2026-04-29
Improve backend DX with hot reload in dev

### LINKED ISSUE #68: Improve backend change DX
if we make a change on the backend we have to restart all the containers b/c our Dockerfile rebuilds the project everytime

there should be some solution (ex. using the go `air` packge?) to hot reload. we should make a separate `Dockerfile.dev` to support this

since we don't have prod deployements, rename the compose to `docker-compose.dev.yml` or whatever the pattern is in Clark

trista-chen-29:
Closes #68 

### Context
Backend changes currently require rebuilding the Docker image and restarting containers.

### Changes
* Added `Dockerfile.dev` for local development
* Added `.air.toml` for Go hot reload
* Updated `docker-compose.dev.yml` to run backend with hot reload and mount source code

### Validation
Screenshot 2026-04-29 at 4 01 13 AM
* Edited backend Go files while running
* Confirmed server automatically reloads without rebuilding containers

https://github.com/user-attachments/assets/566a4307-e1c5-49ea-8c9b-c0058ff55538

### DIFF
--- a/.air.toml
+++ b/.air.toml
@@ -1,0 +1,14 @@
+root = "."
+tmp_dir = "tmp"
+
+[build]
+cmd = "go build -o ./tmp/server ./cmd/server"
+bin = "./tmp/server"
+full_bin = "./tmp/server"
+include_ext = ["go"]
+exclude_dir = ["data", "tmp", "vendor", ".git"]
+delay = 500
+stop_on_error = true
+
+[log]
+time = true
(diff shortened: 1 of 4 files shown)


trista-chen-29:
For `Dockerfile.dev`, should I use `golang:1.25` or `golang:1.25-alpine`? cuz alpine has not been working on my computer (I'm not sure if this is just me), so I switched to `golang:1.25` for the hot-reload setup.

steeevin88 (approved):
this works on my end + thanks for the screen recording 

looks good to merge but feel free to add changes to get alpine working on your end

Issue #115, opened 2026-04-30
Move Docker files into /docker


steeevin88:
let's do something similar to what Clark does: https://github.com/SCE-Development/Clark/tree/dev/docker

PR #116, opened 2026-04-30
Moving docker files into /docker like Clark

maernest04:
PR for #115 

Moved docker files into their own /docker folder, similar to how Clark organizes their docker files.


c1ndytat (approved):
🐹

PR #122, opened 2026-05-01
Add protected attendees endpoint for events

### LINKED ISSUE #96: Determine need for public/private event attendees
should event cards show the attendees registered for the event? this could be a privacy issue, we'd need to ensure that when users are registering for an event, that they know their registration will be publically viewable... so maybe this isn't something we need

miiiyyi:
Closes #96 
Adds a backend-only attendees endpoint for events.

This endpoint returns a privacy-safe attendee list and is only accessible to:
- admins with edit access
- users who are already accepted registrants for that event

## Changes
- added `GET /events/:id/attendees` route
- added `GetEventAttendees` handler
- added attendee access check:
  - event edit access OR accepted registration required
- added minimal attendee response shape
  - returns masked display name
  - does not expose email or registration answers
- added Mongo store method to list accepted registrations for an event
- added store interface method for accepted attendee lookup


steeevin88 (approved):
looks fine to me, can you capitalize the PR title before merging 😄

PR #127, opened 2026-05-03
Refactor Mongo Index Initializations 

### LINKED ISSUE #114: Refactor Mongo Index Initializations
we have a few indexes, let's make a helper that calls all of them instead of having them each explicitly declared

let's work on this after #113 which adds another index call to the events collection

codebyemily:
resolves #114

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -26,14 +26,8 @@ func main() {
 	if err := db.Connect(cfg.MongoURI); err != nil {
 		log.Fatalf("Failed to connect to MongoDB: %v", err)
 	}
-	if err := db.InitWaitlistIndexes(); err != nil {
-		log.Fatalf("Failed to initialize waitlist indexes: %v", err)
-	}
-	if err := db.InitRegistrationIndexes(); err != nil {
-		log.Fatalf("Failed to initialize registration indexes: %v", err)
-	}
-	if err := db.InitEventIndexes(); err != nil {
-		log.Fatalf("Failed to initialize event indexes: %v", err)
+	if err := db.InitMongoIndexes(); err != nil {
+		log.Fatalf("Failed to initialize MongoDB indexes: %v", err)
 	}
 	defer func() {
 		if err := db.Disconnect(); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/server/main.go
```diff
@@ -26,14 +26,8 @@ func main() {
...
 	}
-	if err := db.InitWaitlistIndexes(); err != nil {
-		log.Fatalf("Failed to initialize waitlist indexes: %v", err)
-	}
-	if err := db.InitRegistrationIndexes(); err != nil {
-		log.Fatalf("Failed to initialize registration indexes: %v", err)
-	}
-	if err := db.InitEventIndexes(); err != nil {
-		log.Fatalf("Failed to initialize event indexes: %v", err)
+	if err := db.InitAllIndexes(); err != nil {
```
steeevin88: can we call this `InitMongoIndexes` instead

pkg/db/indexes.go
```diff
@@ -0,0 +1,23 @@
+package db
+
+import "fmt"
+
+// InitAllIndexes runs index setup for every collection. Add new Init*Indexes
+// functions to indexInits only; keep per-domain logic in the respective files.
+func InitAllIndexes() error {
+	for _, x := range indexInits {
```
steeevin88: instead of `x` can we use something more useful? ex. `index`, or `idx`

pkg/db/indexes.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+var indexInits = []struct {
+	name string
+	fn   func() error
+}{
+	{"waitlist", InitWaitlistIndexes},
+	{"registration", InitRegistrationIndexes},
+	{"event", InitEventIndexes},
+}
```
steeevin88: can we just move this into the function above? since we don't export this anywhere anyway

steeevin88: and if so, update the comment above the function

steeevin88 (commented):
can you also rename the PR? thanks

steeevin88 (approved):

Issue #119, opened 2026-05-01
Fix /pkg/database


steeevin88:
poor naming, why does https://github.com/SCE-Development/SCEvents/tree/main/pkg/database exist when we have https://github.com/SCE-Development/SCEvents/tree/main/pkg/db

either merge the two together or make a new dir, update all refs please

PR #132, opened 2026-05-04
Fixing the confusion between pkg/db and pkg/database with renaming

maernest04:
# PR for #119 

## Problem
In our code, we had 2 pkg folders with similar names, one called db, the other called database. As a result, there was some confusion as to what their purposes were. The db folder is meant for runtime persistence layer (connections, Redis, and Mongo store APIs) and the database folder was contained the migration helper that backfills missing fields from struct default tags.

## Solution
Rename the database folder to migration folder and keep the db folder unchanged. This way, it's more clear on their purposes in the project.

### DIFF
--- a/cmd/migrate/main.go
+++ b/cmd/migrate/main.go
@@ -7,8 +7,8 @@ import (
 	"strings"
 
 	"github.com/SCE-Development/SCEvents/internal/config"
-	"github.com/SCE-Development/SCEvents/pkg/database"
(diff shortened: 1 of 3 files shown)


c1ndytat (approved):

KhoaNguyen706 (approved):
lotm

steeevin88 (approved):
thanks

PR #133, opened 2026-05-04
Add automatic mongodb field migration runner

### LINKED ISSUE #126: Create routine to auto-migrate MongoDB fields
#62 is cool but requires manual usage... ideally, we just have a routine that runs this script in parallel with our server

this routine shouldn't run super often b/c it's probably a heavy task/script, so pick an interval to run it at (ex. everyday)

miiiyyi:
Closes #126 
## Changes
- added server-side auto-migration background runner
- runs once on server startup
- continues running on a configurable interval
- iterates through `models.MigrationRegistry`
- calls `database.AutoMigrateDefaults(...)` for each registered collection/model

current default interval is 24 hours unless overridden by env

### DIFF
--- a/cmd/server/migrate_runner.go
+++ b/cmd/server/migrate_runner.go
@@ -10,25 +10,25 @@ import (
 	"github.com/SCE-Development/SCEvents/pkg/models"
 )
 
-func runAutoMigrations(ctx context.Context, interval time.Duration) {
-	runOnce := func() {
-		for key, entry := range models.MigrationRegistry {
-			coll := db.Database().Collection(entry.Collection)
+func runMigrationCycle(ctx context.Context) {
+	for key, entry := range models.MigrationRegistry {
+		coll := db.Database().Collection(entry.Collection)
 
-			updated, err := database.AutoMigrateDefaults(ctx, coll, entry.Model)
-			if err != nil {
-				log.Printf("auto-migrate failed for %s (%s): %v", key, entry.Collection, err)
-				continue
-			}
+		updated, err := database.AutoMigrateDefaults(ctx, coll, entry.Model)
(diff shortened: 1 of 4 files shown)


cmd/server/migrate_runner.go
```diff
@@ -0,0 +1,45 @@
...
+			if err != nil {
+				log.Printf("auto-migrate failed for %s (%s): %v", key, entry.Collection, err)
+				continue
+			}
+
+			if updated > 0 {
+				log.Printf("auto-migrate updated %d fields for %s (%s)", updated, key, entry.Collection)
+			}
+		}
+	}
```
steeevin88: should we maybe just make this a separate function? instead of making it inline, up to you

steeevin88 (approved):
should we update the [docker-compose](https://github.com/SCE-Development/SCEvents/blob/main/docker-compose.yml) to enable this by default?

docker-compose.dev.yml
```diff
@@ -48,6 +48,8 @@ services:
       - SERVER_PORT=${SERVER_PORT:-8002}
       - CLIENT_URL=${CLIENT_URL:-http://localhost:3000}
       - CLIENT_API_URL=${CLIENT_API_URL:-http://host.docker.internal:8080}
+      - AUTO_MIGRATE_ENABLED=${AUTO_MIGRATE_ENABLED:-true}
+      - AUTO_MIGRATE_INTERVAL_HOURS=${AUTO_MIGRATE_INTERVAL_HOURS:-24}
```
steeevin88: oops we should also change the prod compose right? 

i suppose we could also just keep it the dev compose file too, it would only run once (on startup)

Issue #118, opened 2026-05-01
Refactor handlers/event.go


steeevin88:
https://github.com/SCE-Development/SCEvents/blob/main/pkg/handlers/event.go

this file is massive lol i think this is bad

maernest04:
Archiving

PR #136, opened 2026-05-04
Refactored the event.go handler's file

maernest04:
# PR for #118 

## Problem
For the handlers/event.go folder, the folder was cluttered with every event API such as event reading and writing, getting attendee information, registration and waitlist, and admin registration dashboard information. This made the file way too big and cluttered.

## Solution
Separate the handlers/event.go file into separate files, using the same package name "handlers" but separating each event APIs into their own file.


c1ndytat (approved):

steeevin88 (commented):
can we make a new `event` folder in here? and then simplify the file names, ex.
```
handler.go
read.go
write.go
registration.go
admin_registration.go
attendee.go
waitlist.go
```

steeevin88 (changes requested):
we should also split the test file too

PR #137, opened 2026-05-04
Handle pending auth

steeevin88:
**CONTEXT**
- bug happened in SCEvents where we would indicate that SCEvents was down to a signed-out user, despite it being up --> moreover, signing in would then display the calendar, and then logging out would still display the calendar
- what's the issue?
	- not sure, but we know that the `/api/Auth/verify 403` returned a 403...
	- thus we have two potential reasons
		1. user's auth was somehow set to `PENDING` --> b/c pending is -1, we would return that the API is down
		2. user's auth token was expired --> again, we return that the API is down

**CHANGES**
- update middleware to handle invalid / pending auth states --> basically allow pass if the user is signed out

### DIFF
--- a/pkg/middleware/auth.go
+++ b/pkg/middleware/auth.go
@@ -108,8 +108,8 @@ func OptionalAuth(clientAPIURL string) gin.HandlerFunc {
 
 		userID, role, accessLevel, err := verifyAuthHeader(authHeader, clientAPIURL)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


c1ndytat (approved):

maernest04 (approved):
Let that man cook

KhoaNguyen706 (approved):
lftm

Issue #130, opened 2026-05-03
Run pipelines on specific changes


steeevin88:
while pipelines are super helpful, it can be annoying if we make a super small non-impactful change (ex. updating the `README.md`) that doesn't need our potentially long pipelines to change

we should make it so that certain pipelines will run ONLY if certain files change... ex. if we change `.go` files in the server, we definitely should run the integration pipeline --> to do this, we can add path-filters

you'll need to determine when each pipeline should run + when they should be skipped. since all pipelines are merge blocking, ensure that if a pipeline is skipped, we still give a success status so it doesn't PR merges

PR #138, opened 2026-05-04
Modifing pipeline checks to only trigger on specific file changes

maernest04:
# PR for #130 

## Problem
Heavy CI (lint, Go build/tests, Docker integration) ran on every PR, even if its simple PR changes like read me's. We should only run CI's on relevant file changes.

## Solution
Add path-based change detection to the SCEvents workflows so those jobs only run when relevant files change, and add always-green status jobs so merge checks still pass when a workflow is intentionally skipped.

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -7,7 +7,28 @@ on:
     branches: [main]
 
 jobs:
+  detect-go-ci-changes:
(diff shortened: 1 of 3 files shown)


steeevin88 (approved):
looks fine to me, fix the typo in the PR title + we can merge

Issue #125, opened 2026-05-03
Remove /internal


steeevin88:
the only thing in `/internal` is `config.go`, let's just move it into `/pkg`

PR #139, opened 2026-05-05
Remove /interval from production Dockerfile

steeevin88:
image

change was introduced in #125


maernest04 (approved):
Why don't astronauts like gravity? It brings them down.

PR #143, opened 2026-05-12
Move registration to async Producer pattern

steeevin88:
**CONTEXT**
- RegisterForEvent handler was doing synchronous seat reservation via Redis and immediate DB acceptance inline
- this meant the HTTP handler was directly coupled to capacity-checking and acceptance logic
- moving to an async producer pattern decouples the handler from business logic and lets a consumer handle seat reservation + acceptance

**CHANGES**
- replace inline TryTakeEventSeat / MarkRegistrationAccepted / ReleaseEventSeat flow with producer.PublishRegistration
- status code changed from 201 Created to 202 Accepted to reflect async processing
- error messages updated to match the new enqueue-based flow

### DIFF
--- a/pkg/handlers/event.go
+++ b/pkg/handlers/event.go
@@ -678,6 +678,16 @@ func (h *EventHandler) RegisterForEvent(producer *registration.Producer) gin.Han
 			return
 		}
 
+		if ev.MaxAttendees != -1 {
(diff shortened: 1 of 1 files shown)


c1ndytat (approved):
scevents stamp time

